PR #123, opened 2015-01-14
Empty out the RestClient log after each run

fatbusinessman:
Without this, the log file builds up forever, with details of every HTTP request Smokey has ever made. Since we'll be capturing log output for anything interesting anyway, this isn't useful to us.

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -2,6 +2,10 @@
 
 set -x
 
+# Clean up untracked files. This is mainly to remove `smokey-rest-client.log`,
+# otherwise it builds up between jobs.
+git clean -fdx
+
 if [ -z $MYTASK ]; then
   MYTASK="test:skyscapenetwork"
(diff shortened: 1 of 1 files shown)


jenkins.sh
```diff
@@ -14,5 +14,8 @@ if [ ! -f .ruby-version ]; then
   export PATH=$(printf $PATH | awk 'BEGIN { RS=":"; ORS=":" } !/rbenv/' | sed 's/:$//')
 fi
 
+# Reset the RestClient log to an empty file
+cat /dev/null >log/smokey-rest-client.log
```
alext: Instead of this, it's probably easiest to do a `git clean -fdx` at the top of the file.  This is what we do for most other projects.

fatbusinessman: Good idea; fixed in cb85adf.

PR #124, opened 2015-02-02
Use an interactive shell when deploying

alexmuller:
Using a non-interactive shell when executing `bundle` uses bundler which is installed as a system gem.

Newly provisioned machines do not have bundler installed as a system gem and need to use rbenv, which requires an interactive login shell to set up.


rjw1:
:+1:

PR #128, opened 2015-03-11
Disable timing check for the performance platform tests

benilovj:
The 'elapsed time' check has been failing regularly across all environments.
This is a known issue and doesn't require immediate action, so it shouldn't
trigger an error alert in nagios.

If the page takes way too long, the nginx timeout would kick in and the return
code wouldn't be 200 (which is asserted on, so that scenario will be detected).

/cc @tombooth

### DIFF
--- a/features/performance_platform.feature
+++ b/features/performance_platform.feature
@@ -8,7 +8,6 @@ Feature: Performance Platform
     When I visit "/performance"
     Then I should get a 200 status code
     And I should see "Performance"
-    And the elapsed time should be less than 5 seconds
 
   @high
   Scenario: Performance Platform dashboards are available
@@ -18,4 +17,3 @@ Feature: Performance Platform
     When I visit "/performance/carers-allowance"
     Then I should get a 200 status code
     And I should see "Carer's Allowance"
-    And the elapsed time should be less than 5 seconds


bradleywright:
How long do these pages take now? Presumably to succeed it's between 5 and 10 seconds.

benilovj:
@bradleywright we've just looked through some of the past Smokey test failures. There's been a couple of times where it's taken about 8 seconds, but also odd case where it's taken 32s (network problem?). There's undoubtedly a more robust way of looking at these timings (eg in the PP kibana).

However, the fact that the PP team is aware of this and it isn't being fixed imminently, implies to me that we shouldn't be raising an alert for it.

alext:
Given that the PP team own this endpoint, maybe this check should be moved to [pp-smokey](https://github.com/alphagov/pp-smokey).

bradleywright:
These URLs live in Spotlight, so all we're really asserting is that we have an active route for `/performance`. I'm not sure that's worth between 15 and 20 seconds.

Also, [since Spotlight is already tested in `pp-smokey`](https://github.com/alphagov/pp-smokey/blob/master/features/spotlight.feature) I think I agree with @alext .

benilovj:
@bradleywright so getting rid of this feature completely then?

bradleywright:
@benilovj assuming @tombooth or similar agrees, I think so.

benilovj:
@tombooth has given his blessing on slack to remove the feature. I'll close this PR and raise another one with the deletion.

PR #129, opened 2015-03-11
Remove the performance platform feature

benilovj:
- [Spotlight is already being tested in `pp-smokey`](https://github.com/alphagov/pp-smokey/blob/master/features/spotlight.feature)
- when these scenarios fail, there isn't an immediate course of action that the 2nd line GOV.UK support team can take to rectify the failures
- these tests fail regularly because they are sensitive to performance changes to the `/performance` pages

/cc @tombooth @bradleywright


bradleywright:
:+1:

mattbostock:
We should remove the Icinga [check](https://github.gds/gds/puppet/blob/05723ac57a647aa1ebfcf5773c676a84bbfdee1e/modules/monitoring/manifests/checks/smokey.pp#L31) first.

benilovj:
The Icinga check has been removed in production, this is safe to merge.

PR #130, opened 2015-04-08
Update smartanswers test to step through a different answer

tadast:
As can be checked on preview by visiting `/student-finance-calculator/y/2013-2014/full-time`, the answer kept redirecting to the first question, because the dates in the URL are no longer valid as answers. Therefore the test was not testing the flow as intended.

This picks another answer to test against that is less likely to change the possible answers.

### DIFF
--- a/features/smartanswers.feature
+++ b/features/smartanswers.feature
@@ -14,14 +14,15 @@ Feature: Smart Answers
       | /overseas-passports                         |
       | /pay-leave-for-parents                      |
       | /register-a-death                           |
+      | /vat-payment-deadlines                      |
 
-    @normal
-    Scenario: step through a smart answer
-      Given I am testing through the full stack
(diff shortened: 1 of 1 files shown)


ajlanghorn:
@tadast Is a date fifteen years ago likely to stick around for much longer? For future-proofing, would it not be better using 2015-01-31 or similar?

tadast:
@ajlanghorn I chose this example because, unlike other calculators, this one does not expire and does not need any date-dependant configuration, so my bet is it's gonna stick around for a long time :) It may happen that a `cheque` option will disappear though, but I'm not sure how to prevent those things from happening other than having a "fake smart answer" for smokey tests only, but that's bad too.

ajlanghorn:
@tadast Okay, that sounds reasonable. :penguin:

PR #136, opened 2015-05-27
Add test for trailing slash behaviour

alexmuller:
This behaviour is implemented in our Varnish VCL and is not currently tested anywhere.

https://github.gds/gds/puppet/blob/06b39b70339/modules/varnish/templates/default.vcl.erb#L31-37


mattbostock:
:+1:

PR #138, opened 2015-06-02
Add signon cookie tests

alexmuller:
These cookies need to be secure because they contain authentication
information. We set the HttpOnly flag because there's no reason for
client-side script to access the cookie.

My intention with the regular expression is to match the following:

- A delimiter in the Set-Cookie header (semicolon)
- Then a space
- Then the name of the property we want to find
- Then either another delimiter (semicolon) or the end of the header

### DIFF
--- a/features/step_definitions/cookie_steps.rb
+++ b/features/step_definitions/cookie_steps.rb
@@ -1,0 +1,7 @@
+Then /^I should receive a "([a-z0-9_]+)" cookie which is "([a-zA-Z]+)"$/ do |cookie_name, cookie_property|
+  header = page.response_headers['set-cookie']
+  assert header.start_with?(cookie_name), "No cookie called #{cookie_name} is being set"
+
+  property_matches = header.match /; #{cookie_property}(;|$)/
+  assert !property_matches.nil?, "The cookie #{cookie_name} does not have property #{cookie_property}"
+end
--- a/features/signon.feature
+++ b/features/signon.feature
@@ -4,3 +4,8 @@ Feature: Sign-on-o-tron
     When I try to login as a user
     Then I should see "Welcome to GOV.UK"
 
+  @low
+  Scenario: Signon cookies are marked as secure and HttpOnly
+    When I go to the "signon" landing page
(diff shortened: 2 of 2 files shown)


features/step_definitions/cookie_steps.rb
```diff
@@ -0,0 +1,10 @@
+Then /^I should receive a "([a-z0-9_]+)" cookie which is "([a-zA-Z]+)"$/ do |cookie_name, cookie_property|
+  headers = @response.headers[:set_cookie].select do |h|
```
bradleywright: Can you use `response.cookies` here?

https://github.com/rest-client/rest-client#cookies

Might remove some of the funky regular expression stuff.

features/step_definitions/smokey_steps.rb
```diff
@@ -31,7 +31,7 @@
 
   page.driver.browser.agent.add_auth(base_host, ENV['AUTH_USERNAME'], ENV['AUTH_PASSWORD'])
 
-  visit url
+  @response = get_request(url, default_request_options)
```
alext: This breaks several of the other tests (eg `mainstream_publishing_tools.feature`).  `visit` is a capybara method that makes a request, and then sets up the response on the `page` variable.  Later steps then rely on this being populated to make assertions against etc.  Swapping this out for a basic rest-client request causes all the later assertions to fail.

alexmuller: Thanks, I find it really difficult to figure out how (our usage of?) Cucumber/Capybara/rest-client etc all tie together. I've removed this commit and started using a `Capybara::Session` instead.

features/step_definitions/cookie_steps.rb
alexmuller: I've stopped using rest-client so I'll see if there's a nicer way to do this with `Capybara::Session`. Maybe [show_me_the_cookies](https://github.com/nruth/show_me_the_cookies).

surminus:
Me and @ajlanghorn checked this, looks good and test passed, happy to merge :)

PR #141, opened 2015-06-18
Draft smokey features

tekin:
This adds a rake task for running smokey features for the draft stacks: 

```
$ bundle exec rake test:preview_draft
```

and

```
$ bundle exec rake test:draft
```

Features that apply to the draft stack should be marked with the `@draft` tag.

It's worth noting that we do not currently have any features for either static or manuals-frontend.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,6 +6,16 @@ Cucumber::Rake::Task.new("test:preview",
   t.cucumber_opts = %w{--format progress -t ~@pending -t ~@notpreview}
 end
 
+Cucumber::Rake::Task.new("test:draft",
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -6,6 +6,16 @@ Cucumber::Rake::Task.new("test:preview",
   t.cucumber_opts = %w{--format progress -t ~@pending -t ~@notpreview}
 end
 
+Cucumber::Rake::Task.new("test:draft",
+    "Run all tests that are valid in our preview environment") do |t|
```
danielroseman: draft

PR #149, opened 2015-08-10
Assign priorities to draft environment checks

heathd:
in order to have the smokey checks verified through nagios (and thus
trigger monitoring alerts), they need to be assigned priorities.

### DIFF
--- a/features/draft_environment.feature
+++ b/features/draft_environment.feature
@@ -3,7 +3,7 @@ Feature: Draft environment
   GOV.UK. Accessing the draft environment requires a valid signon session.
   Access to the draft stack should be denied without a valid signon session.
 
-  @draft
+  @draft @high
   Scenario: visiting a draft page requires a signon session
     When I attempt to go to a case study
     Then I should be prompted to log in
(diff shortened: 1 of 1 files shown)


heathd:
checked this priority assignment in slack with @alexmuller who felt that they were reasonable

features/draft_environment.feature
```diff
@@ -11,14 +11,14 @@ Feature: Draft environment
     Then I should be on the case study page
     And the page should contain the draft watermark
 
-  @draft
+  @draft @high
```
bradleywright: Not sure this is a page we want triggered by Nagios. We had issues with named pages in the Whitehall smoke tests (they'd get unpublished etc.).

Is there a way we could dynamically fetch a page?

heathd: hmm, there's no api in content store to expose searching/filtering by `format`, so I can't think of a way of doing that, unless the smokey tests would ssh onto a `backend` machine and interrogate the specialist publisher database directly.

Maybe it's a case of 'this will do for now', and if/when the document disappears we can try to make the test more robust, or remove it from the monitoring checks. By that point the publishing API work may be more advanced and we may have more options for addressing this.

Open to suggestions if anyone has any ideas.

heathd: @danielroseman suggested that we could pick a document using content-register. That seems like a viable option although there's no option in content-register to limit the number of results it returns, so basically we'd need to query and get back a list of all 1200 case studies and then pick one (351kb, request takes 0.5 seconds). 

Let me know if you think this is worth doing @bradleywright?

bradleywright: I would just remove the Nagios check for this. It's valuable to know that Smokey has failed, but we shouldn't be notifying the alerting system.

To make this check more useful for Smokey we should also do what you suggest with content-register.

heathd: ok thanks, do you agree with the priorities assigned to the other checks in this file? and would you mind also reviewing the ones in #150? Thanks

PR #150, opened 2015-08-10
Add tests for specialist publisher formats on draft

heathd:
These check that the specialist publisher format documents can be loaded
in the draft environment.

The URLs chosen are arbitrary ones which hopefully should be in
existence.


elliotcm:
The implementation looks fine but I'm not comfortable determining whether this is appropriate for triggering high priority alerts.

heathd:
after discussion with Brad, I've removed the triggering of nagios alerts for these tests

PR #151, opened 2015-08-11
Special case for production draft url

heathd:
smokey incorrectly guesses the draft url in production.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -2,15 +2,10 @@ require 'nokogiri'
 require 'capybara/cucumber'
 require 'capybara/mechanize'
 require 'uri'
-
-def guess_draft_url_from_live(live_url)
-  url = URI.parse(live_url)
-  host = (["draft-origin"] + url.host.split(".")[1..-1]).join(".")
-  "#{url.scheme}://#{host}"
-end
+require 'plek'
 
 ENV["GOVUK_WEBSITE_ROOT"] ||= "https://www.preview.alphagov.co.uk"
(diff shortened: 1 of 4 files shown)


bradleywright:
This needs a FIXME as the fix will stop working when we switch to publishing.service for production.

features/support/env.rb
```diff
@@ -4,9 +4,13 @@
...
-  url = URI.parse(live_url)
-  host = (["draft-origin"] + url.host.split(".")[1..-1]).join(".")
-  "#{url.scheme}://#{host}"
+  if live_url == 'https://www.gov.uk'
+    'https://draft-origin.production.alphagov.co.uk'
+  else
+    url = URI.parse(live_url)
+    host = (["draft-origin"] + url.host.split(".")[1..-1]).join(".")
+    "#{url.scheme}://#{host}"
+  end
```
alext: Is there a reason this can't use `Plek.find('draft-origin')`?

mattbostock:
I'd rather avoid a FIXME if we can as it's another thing we'll need to worry about; if we could avoid hardcoding this, that would be ideal.

features/support/env.rb
bradleywright: Yeah, actually that's better as it'll always work and be portable.

heathd:
good suggestion, thanks. I've changed it to use plek.

mattbostock:
:+1:

bradleywright:
:+1:

heathd:
Smokey has now passed in production https://deploy.production.alphagov.co.uk/job/Smokey_Production/5302/console

PR #152, opened 2015-09-03
Fix licensify hostname

alexmuller:
Use GOVUK_APP_DOMAIN because it's more portable than hardcoding a hostname.


mattbostock:
:+1:

PR #153, opened 2015-09-14
Update smokey to check topics dynamically

mobaig:
Add additional step to traverse topic hierarchy across both first and
second levels, to improve monitoring of /topic path.

Story: https://trello.com/c/Dj4P9D8N

### DIFF
--- a/features/step_definitions/smokey_steps.rb
+++ b/features/step_definitions/smokey_steps.rb
@@ -169,3 +169,10 @@ end
 Then /^I should see Publisher's publication index$/ do
   page.should have_selector("#publication-list-container")
 end
+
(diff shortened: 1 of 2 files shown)


rboulton:
I think this will fail if there's a top-level topic which has no subtopics.  This is sometimes the case while new topics are being built - I'm not sure we want a smokey alert because of it.

It would be nice if this checked a larger number of topic pages than 1 (at each level).  It feels likely to miss issues which occur on a subset of pages (such as the recent incident where browse pages were broken, but only if they didn't have links to topic pages).

mobaig:
closing and continuing on #154

PR #154, opened 2015-09-14
Update smokey to check browse and topics dynamically

mobaig:
Add additional step to traverse browse categories across both first and
second levels, to improve monitoring of /browse path.

Stories: https://trello.com/c/VqFDnMot and https://trello.com/c/Dj4P9D8N

### DIFF
--- a/features/topics.feature
+++ b/features/topics.feature
@@ -1,0 +1,9 @@
+Feature: Topics
+
+  Scenario: dynamically checking topic hierarchy
+    Given I am testing through the full stack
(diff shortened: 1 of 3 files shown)


bishboria:
:+1:

features/topics.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Topics
+
+  Scenario: dynamically checking topic hierarchy
```
alext: This needs a tag so that it gets run by smokey (probably `@high`)

PR #156, opened 2015-09-28
[#104329818] Add monitoring for GOVUK_WEBSITE_ROOT

alexmuller:
We had an incident recently where we accidentally changed the content of travel advice feeds which caused GovDelivery to send email alerts when there were no actual updates.

This commit tests that the GOVUK_WEBSITE_ROOT environment variable is configured correctly on our machines. In preview and staging the expected value is always equal to the actual value. This test is only valuable in production where it makes sure that the website root is set to `https://www.gov.uk`.

I've also had to add a new step definition to include the value of the environment variable which I'm not particularly happy about but it's the best I've come up with so far.

---

I'm not particularly happy with any of this code which is probably an indication that it's the wrong approach to be taking. If anyone has any better ideas for monitoring that the website root is set correctly in production I would really like to hear them.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -7,5 +7,12 @@ require 'plek'
 ENV["GOVUK_WEBSITE_ROOT"] ||= "https://www.preview.alphagov.co.uk"
 ENV["GOVUK_DRAFT_WEBSITE_ROOT"] ||= Plek.find('draft-origin')
 
+case ENV["GOVUK_WEBSITE_ROOT"]
+when "https://www.preview.alphagov.co.uk", "https://www-origin.staging.publishing.service.gov.uk"
+  ENV["EXPECTED_GOVUK_WEBSITE_ROOT"] = ENV["GOVUK_WEBSITE_ROOT"]
+else
+  ENV["EXPECTED_GOVUK_WEBSITE_ROOT"] = 'https://www.gov.uk'
+end
+
 Capybara.default_driver = :mechanize
 Capybara.app_host = ENV["GOVUK_WEBSITE_ROOT"]
--- a/features/travel_advice.feature
+++ b/features/travel_advice.feature
@@ -23,3 +23,8 @@ Feature: Travel Advice
       | Path                                   |
       | /foreign-travel-advice.atom            |
       | /foreign-travel-advice/luxembourg.atom |
+
+  @normal
+  Scenario: Country feed should contain correct website root
+    When I visit "/foreign-travel-advice/ireland.atom"
+    Then the XML ID is formed from the correct URL
--- a/features/gov_uk.feature
+++ b/features/gov_uk.feature
@@ -5,3 +5,10 @@ Feature: Core GOV.UK behaviour
     When I visit "https://www.gov.uk/browse/benefits/" without following redirects
     Then I should get a 301 status code
     And I should get a location of "/browse/benefits"
+
+  @normal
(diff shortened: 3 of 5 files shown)


features/travel_advice.feature
```diff
@@ -23,3 +23,8 @@ Feature: Travel Advice
       | Path                                   |
       | /foreign-travel-advice.atom            |
       | /foreign-travel-advice/luxembourg.atom |
+
+  @normal
+  Scenario: Country feed should contain correct website root
+    When I visit "/foreign-travel-advice/ireland.atom"
+    Then I should see the contents of environment variable "EXPECTED_GOVUK_WEBSITE_ROOT" followed by "/foreign-travel-advice/ireland"
```
rjw1: in this case dont we want to make sure that the atom id is correct since that is what triggered the bug. 
```
<id>https://www.gov.uk/foreign-travel-advice/ireland#2015-09-11T11:56:59+01:00</id>
```

alexmuller: You're right, I was being lazy. Updated.

features/support/env.rb
```diff
@@ -7,5 +7,13 @@
 ENV["GOVUK_WEBSITE_ROOT"] ||= "https://www.preview.alphagov.co.uk"
 ENV["GOVUK_DRAFT_WEBSITE_ROOT"] ||= Plek.find('draft-origin')
 
+if ENV["GOVUK_WEBSITE_ROOT"] == "https://www.preview.alphagov.co.uk"
+  ENV["EXPECTED_GOVUK_WEBSITE_ROOT"] = ENV["GOVUK_WEBSITE_ROOT"]
+elsif ENV["GOVUK_WEBSITE_ROOT"] == "https://www-origin.staging.publishing.service.gov.uk"
+  ENV["EXPECTED_GOVUK_WEBSITE_ROOT"] = ENV["GOVUK_WEBSITE_ROOT"]
+else
+  ENV["EXPECTED_GOVUK_WEBSITE_ROOT"] = 'https://www.gov.uk'
+end
```
mattbostock: A case statement would be cleaner here.

mattbostock:
Agree that this doesn't feel ideal, especially us hardcoding the Preview and Staging URLs, though this should be sufficient for what we need and I don't have a better method to suggest.

:+1: happy to merge as-is, though please see my comment above about using a case statement.

features/support/env.rb
alexmuller: Agreed, updated.

mattbostock:
I think it'd be worth explaining in 808aa26 that this test is a useful canary to see if we've correctly configured the website root.

alexmuller:
Sure, updated. Also made the steps less generic based on feedback from @alext.

features/travel_advice.feature
```diff
@@ -23,3 +23,8 @@ Feature: Travel Advice
       | Path                                   |
       | /foreign-travel-advice.atom            |
       | /foreign-travel-advice/luxembourg.atom |
+
+  @normal
+  Scenario: Country feed should contain correct website root
+    When I visit "/foreign-travel-advice/ireland.atom"
+    Then the XML ID should be correct
```
mattbostock: 'should be correct' could be more specific, e.g. 'links to the correct website root'?

alexmuller: Yep, updated to be more specific.

mattbostock:
:+1:

PR #157, opened 2015-10-02
Remove logging for /browse & /topic test

tijmenb:
Smokey currently outputs the current page when traversing browse and topic pages. This makes the log output a bit noisy in CI.

@mobaig agree?


mobaig:
@tijmenb fine with this change, but then I'm not a consumer of this info. If it's not used by anyone, nuke it!

tijmenb:
Not used! Can you merge?

PR #163, opened 2015-11-25
Check an announcement from page 2 of the list

boffbowsh:
In staging, rummager is often out of sync with the whitehall DB, and so
the top announcement may not exist in Whitehall itself, resulting in a
404. Using the top announcement from page 2 is a lot more likely to
succeed.

### DIFF
--- a/features/step_definitions/whitehall_steps.rb
+++ b/features/step_definitions/whitehall_steps.rb
@@ -13,7 +13,7 @@ Then /^I should be able to view publications$/ do
 end
 
 Then /^I should be able to view announcements$/ do
-  follow_link_to_first_announcement_on_announcements_page
+  follow_link_to_first_announcement_on_announcements_page(2)
 end
 
 When /^I do a whitehall search for "([^"]*)"$/ do |term|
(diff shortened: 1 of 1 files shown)


elliotcm:
You could also check the last one on the first page and not encode our pagination in the tests, but sure.

bradwright:
Don't agree with this change. Smoke tests are meant to emulate some high level user journeys, and clicking through to the 2nd page isn't one of them.

If the sync fails this often we should be fixing that, not adding hacks into our test suite.

elliotcm:
@bradwright Do you want to revert?

boffbowsh:
The issue isn't that sync fails, it's that (I think) the restored elasticsearch index is out of sync slightly with the Whitehall DB. I can look into fixing this but I'm not sure how. Good point though, and happy for this to be reverted if needed.

bradwright:
@boffbowsh sure, I was using `sync fails` as lazy shorthand for the entire thing. Let's not revert it for the moment in order to get things working, but leaving this in permanently isn't something we should do.

boffbowsh:
Raised in 2nd line backlog https://trello.com/c/jucvPJGF/73-fix-sync-between-rummager-and-whitehall-in-staging

PR #164, opened 2016-01-05
Replace 'preview' references with 'integration'

benilovj:
The preview environment has been retired and replaced by the
integration environment in Dec 2015.

/cc @jamiecobbett


features/support/base_urls.rb
```diff
@@ -1,5 +1,5 @@
 def app_domain
-  ENV["GOVUK_APP_DOMAIN"] || "preview.alphagov.co.uk"
+  ENV["GOVUK_APP_DOMAIN"] || "integration.alphagov.co.uk"
```
alexmuller: This should be a `service.gov.uk` subdomain

benilovj: oops – search-and-replace error. Fixed and force pushed.

PR #168, opened 2016-02-10
Update frontend features

davidbasalla:

This is an attempt to make tests for entering a postcode less brittle by moving away from searching for the string of the label "Enter a postcode" and towards searching for the input field by class. That enables us to change the label text in the `frontend` app without breaking the smokey tests.

Additionally this PR attempts to make the `frontend` tests more accurate by specifying HTML elements as well as text when searching for an element on a page.

### DIFF
--- a/features/step_definitions/frontend_steps.rb
+++ b/features/step_definitions/frontend_steps.rb
@@ -9,3 +9,7 @@ Then /^I should see the services and information section on the homepage$/ do
   doc = Nokogiri::HTML(html)
   assert doc.css('#services-and-information')
 end
+
+Then /^I should see an input field for entering my postcode$/ do
+  @response.body.should have_field('postcode')
+end
--- a/features/step_definitions/frontend_steps.rb
+++ b/features/step_definitions/frontend_steps.rb
@@ -13,3 +13,15 @@ end
 Then /^I should see an input field for entering my postcode$/ do
   @response.body.should have_field('postcode')
 end
+
+Then /^I should see the heading "(.*?)"$/ do |title|
+  @response.body.should have_css('h1', text: title)
+end
+
+Then /^I should see the matched local authority "(.*?)"$/ do |local_authority|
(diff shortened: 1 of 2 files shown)


features/step_definitions/frontend_steps.rb
```diff
@@ -9,3 +9,19 @@
   doc = Nokogiri::HTML(html)
   assert doc.css('#services-and-information')
 end
+
+Then /^I should see an input field for entering my postcode$/ do
+  @response.body.should have_field('postcode')
+end
+
+Then /^I should see the heading "(.*?)"$/ do |title|
+  @response.body.should have_css('h1', text: title)
```
sihugh: Does this do an exact match on text?  Not sure whether the UK passport test (for example) would fail as the header only contains that text and is not an exact match.

davidbasalla: Hey, I've traced the method down to the matcher and it seems to run [`node.text.match(options[:text])`](https://github.com/jnicklas/capybara/blob/e7dc237ead86d1da826b2dfc92c6f2af8a21d7bb/lib/capybara/node/finders.rb#L182), using the `String` object's `match` method. Meaning the searched for text only has to be present in order for it to match, rather than be an exact match. So kind of like an include..?

tijmenb:
I'm unsure about this. You could say that relying on css selectors will make smokey test the implementation, not the behaviour, of the apps (a user won't care about the existence of a specific element, just if whether it works). I suspect smokey is only meant to be super-high level.

Perhaps changing the tests from the copy to the HTML is also just moving the brittleness from one place to another. 

I would argue that the best solution for the copy brittleness in https://github.com/alphagov/smokey/pull/167/files is to remove the `And I should see "Enter a postcode"` assertion entirely. If the app would break, the next assertion would still raise an error.

davidbasalla:
Closed in favour of https://github.com/alphagov/smokey/pull/169

PR #169, opened 2016-02-12
Remove check for enter postcode instruction

davidbasalla:

This assertion is not required as the next assertion tests that the postcode form is present, and removing it allows us to change the "Enter a postcode" instruction text in the `frontend` app without having to update smokey tests.

The change is motivated to make the test less brittle. Initially we attempted to test with css selectors, but this was testing implementation which the individual apps should already do. See https://github.com/alphagov/smokey/pull/168

### DIFF
--- a/features/frontend.feature
+++ b/features/frontend.feature
@@ -44,7 +44,6 @@ Feature: Frontend
   Scenario: check licences load
     When I visit "/busking-licence"
     Then I should see "Busking licence"
-     And I should see "Enter a postcode"
     When I try to post to "/busking-licence" with "postcode=E20+2ST"
     Then I should get a 200 status code
      And I should see "Busking licence"


jennyd:
This test was [added fairly recently](https://github.com/alphagov/smokey/pull/161/files) - you could talk to @alexmuller about the motivation for adding both the GET and POST cases.

alexmuller:
We added this check because of a failure that we didn't notice when migrating between environments. It's possible that the text check duplicates the `POST check`. The steps I'd like to see to validate this are:

- Switch off Licensify in integration (either all the machines, or just stop the apps)
- Verify that this test fails (and which step it fails on)
- [Deploy this branch](https://deploy.integration.publishing.service.gov.uk/job/Smokey_Deploy/) (might require some devopsing)
- Verify that this test still fails (and which step)

Happy to help with testing this if you need a hand.

I'd still be a little concerned that we wouldn't be testing that we're actually displaying a form on the frontend that users can put their postcode in. Is there a way we can test that which is less intrusive?

boffbowsh:
I'm starting to feel like the amount of back-and-forth that this change has required outweighs the amount of time that'd be needed to change the text in Smokey if the frontend text changes. There must be other examples of slightly brittle tests in Smokey that would break if we changed frontend copy, layouts etc?

h-lame:
Given the next step is to post with the querystring `postcode=E20+2ST` would it be enough to check that there is an input field that targets the `postcode` param on the page?  Should be less brittle than checking for the text in the label, but also it doesn't test the implementation as much as the css approach in #168 did.

It does make me wonder why smokey doesn't use the UI though and does a direct POST.

boffbowsh:
`form input[name=postcode]` would be good :+1:

davidbasalla:
Closed in favor of https://github.com/alphagov/smokey/pull/170

alexmuller:
Checking that the input is correct is better :+1:

(though as @h-lame noticed I should really have just done the POST through the interface like we do for the [signon steps](https://github.com/alphagov/smokey/blob/a7f39ddfd4455b4502687754aed50e5a7ba28e64/features/step_definitions/signon_steps.rb))

davidbasalla:
@alexmuller Maybe you didn't do it because in this case [the `visit` step assigns to `@response`](https://github.com/alphagov/smokey/blob/master/features/step_definitions/smokey_steps.rb#L47-L49), so methods like `fill_in` and `click_button` that work with `page` aren't directly available...

PR #170, opened 2016-02-15
Update postcode form step to check for input element

davidbasalla:

This is an attempt to make tests for entering a postcode less brittle by moving away from searching for the string of the label "Enter a postcode" and towards searching for the input field by class. That enables us to change the label text in the frontend app without breaking the smokey tests.

See discussion in https://github.com/alphagov/smokey/pull/169


issyl0:
LGTM, but I wasn't involved at all in the previous discussion about the approach.

h-lame:
Seems like a good compromise between all discussed approaches.  Merging.

PR #172, opened 2016-03-14
Add simple feature for service manual guide and topic pages

AndrewVos:


features/step_definitions/smokey_steps.rb
```diff
@@ -135,9 +135,9 @@ def should_see(text)
 
 Then /^I should see "(.*)"$/ do |content|
   if @response
-    @response.body.include?(content).should == true
+    expect(@response.body).to include(content)
   elsif page
-    page.body.include?(content).should == true
+    expect(page.body).to include(content)
```
alexmuller: What's the reason for changing the syntax of the assertions?

AndrewVos: @alexmuller I was getting deprecation warnings when running my tests. We should go through and replace everything at some point.

PR #173, opened 2016-03-16
Revert "Upgrade ruby to 2.3"

alexmuller:
This reverts commit d5a578e9c0fe925e2adfcfee833acfa242a23ae4.

This commit broke one of our tests because of the upgrade of rest-client from 1.6.7 to 1.8.0, which introduces lots of changes to the way SSL is handled.

It's more important to have working monitoring than to upgrade all the gems so we should do this upgrade more cautiously and in smaller chunks.


deanwilson:
Looks fine and we need the monitoring coverage back.

PR #174, opened 2016-03-30
Try upgrading to ruby 2.3 again

benilovj:
This was previously attempted in #171 and reverted in #173. The difference in this attempt is that `rest-client` is not updated.

/cc @AndrewVos @alexmuller


AndrewVos:
Me reviewing this feels redundant but LGTM :smile:

benilovj:
The test suite is passing on staging /cc @rjw1

PR #175, opened 2016-04-15
Update licensify-admin Smokey search term

surminus:
Smokey tests started to fail for licensify-admin for reasons not quite yet determined. Upon testing manually, the assigned user is able to login but is not presented with any applications. This appears to be
correct behaviour, but the search term that is specified is not present. This commit updates it to a specific term when that user logs in.

/cc @alexmuller who added the test in f5e48da4 to ensure that what we're seeing _is_ the correct behaviour.

### DIFF
--- a/features/licensing.feature
+++ b/features/licensing.feature
@@ -24,4 +24,4 @@ Feature: Licensing
   Scenario: Signing in to licensify-admin
      When I try to login as a user
       And I login to Licensify
-     Then I should see "Tacit?"
+     Then I should see "There are no applications available for processing"


deanwilson:
The code change looks fine but I'd agree with double checking with The Muller.

surminus:
Muller

alexmuller:
I'm a bit confused about what's changed to make this happen, I'll have a look around quickly.

alexmuller:
@surminus: Would you mind updating the commit message with the context from the email David sent us?

surminus:
@alexmuller, let me know if it makes sense

alexmuller:
Perfect, I think this is good to merge :+1:

PR #176, opened 2016-04-27
Change search term for Smokey licensing test

surminus:
Someone added a test license so the search term changed again. Changed the search term to "Sign Out" which is present if one is logged in, but _shouldn't_ change if test licenses are added (and/or if there is a difference between Staging and Production).


alexmuller:
I'm not sure that this is the right test - the one that's been modified here is for the frontend but the one that's failing is for the admin app.

surminus:
@alexmuller you're completely right, f$%&

PR #182, opened 2016-05-17
Fix logging of errors.

danielroseman:
syslog chokes with non-ASCII characters (returning a very unhelpful
"TypeError: [priority,] message string"). But on a failing test, we log
the entire page output, which includes the (non-ASCII) copyright symbol.

Encode messages to ASCII with the 'xmlcharrefreplace' codec, which
replaces the values with their XML-escaped equivalents.

### DIFF
--- a/nagios_check_cache.py
+++ b/nagios_check_cache.py
@@ -57,9 +57,10 @@ class FeatureTestRun(object):
 
     def write_step_description_and_status(self, result_details, feature, priority, scenario, step):
         result_details.add("    Step: [%s] %s%s\n" % ((self.step_status(step)), step['keyword'], step['name']))
-        syslog(LOG_NOTICE, "%s | %s%s | %s | %s%s" % (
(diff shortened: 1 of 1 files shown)


alexmuller:
@MatMoore and @brenetic might be interested in this because it happened to us on 2nd line too.

MatMoore:
This looks related but it's not the same problem we looked at. In our case, I looked a JSON report created by cucumber when the tests are run. At the time, it was not able to render a JSON report at all, due to some character encoding issue - so the only reason the python code failed was because the file was empty.

We didn't identify the source of the problem, but it only affected particular versions of our dependencies, so I reverted them back to the versions we we're using before the ruby upgrade.

This commit looks fine to me, but I would expect `message.encode('utf8')` to also work. Not sure if the intent was just to get rid of the problem characters or actually produce some XML though.

More generally in python 2.x I would use the unicode string type (e.g. u'hello world' for string literals) for any string manipulation and only encode at the end.

PR #186, opened 2016-06-22
Add downcase feature

gpeng:
Functionality has been added to the router LB that downcases URLs that are entirely in capitals. This PR adds checks that:

* Entirely upper case urls are 301 redirected to their lowercase equivalent
* Partially upper case urls are ignored

Part of [Trello](https://trello.com/c/CDpTRF1e/408-404s-caused-by-capitalisation-medium)

### DIFF
--- a/features/gov_uk.feature
+++ b/features/gov_uk.feature
@@ -12,3 +12,16 @@ Feature: Core GOV.UK behaviour
     And I force a varnish cache miss
     When I visit "/"
     Then the logo should link to the homepage
+
+  @normal
+  Scenario: entirely upper case slugs redirect to lowercase
+    Given I visit "/GOVERNMENT/PUBLICATIONS" without following redirects
+    Then I should get a 301 status code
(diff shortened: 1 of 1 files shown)


boffbowsh:
Merge when https://github.com/alphagov/govuk-puppet/pull/4524 is deployed

features/gov_uk.feature
```diff
@@ -12,3 +12,16 @@ Feature: Core GOV.UK behaviour
     And I force a varnish cache miss
     When I visit "/"
     Then the logo should link to the homepage
+
+  @normal
+  Scenario: entirely upper case slugs redirect to lowercase
+    Given I visit "https://www.gov.uk/GOVERNMENT/PUBLICATIONS" without following redirects
```
alexmuller: I think these tests should both drop the hostname and start with a leading slash so that they test the right thing in integration and staging.

gpeng: Fixed in 7fadda5

PR #189, opened 2016-07-05
Add local-links-manager base URL

jennyd:
This is needed for the test added in 05394a7d081a2ae759c9215aef0d3baf2300b79e.


sihugh:
That does seem like a thing it would need

PR #190, opened 2016-07-05
Gem updates

dougdroper:


alexmuller:
We've had lots of problems when upgrading all the gems in past, eg #171 #173 #174 #179 which makes me a bit nervous about this change.

dougdroper:
Hmm interesting. I guess there are a lot of test failures already, and this is easy to revert. It would be good to run tests against a branch to make sure that it doesn't just run on my machine, but I am happy to revert this if it does cause issues.

alexmuller:
I think I'd be more comfortable with this if I understood what was changing and why - right now it looks like we're just upgrading everything because that will make it work rather than fixing specific problems.

features/licensing.feature
```diff
@@ -8,19 +8,17 @@ Feature: Licensing
...
     Given I am testing "licensing"
       And I am benchmarking
       And I am testing through the full stack
       And I force a varnish cache miss
-    When I visit "/apply-for-a-licence/forms/bury/test-licence/9999-7-1,0-1"
+    When I visit "/apply-for-a-licence/test-licence/westminster/apply-1"
     Then the elapsed time should be less than 10 seconds
 
-  @normal
+  @normal @pending
```
alexmuller: This should almost certainly be working, do we know what's wrong with it?

dougdroper:
> what was changing and why
yeah the upgrade was to just get rid of the deprecation notice, not to fix anything specific. I think I will split it out into separate commits so that it is clearer what fixes some of the tests.

features/licensing.feature
dougdroper: I couldn't get the page to load at all, will remove the pending since it doesn't really solve anything.

dougdroper: Also I couldn't get:
/apply-for-a-licence/test-licence/westminster/apply-1/form
/apply-for-a-licence/forms/bury/test-licence/9999-7-1,0-1

to load either?

dougdroper:
closing in favour of: https://github.com/alphagov/smokey/pull/194

PR #193, opened 2016-07-14
Remove Brad

jackscotti:


alexmuller:
This is interesting. Does @daibach know about this test? I've never seen it before. It'd be good to avoid these kinds of comments and actually explain what's going on if we can.

alexmuller:
Oh, I see it was pushed directly to master in October 2015 with no explanation in the commit.

jackscotti:
I asked him whether adding his name there would have been fine.

alexmuller:
Cool. I think this change is fine in general, but I dislike the principle of having stuff in our codebase which just says "ask [person]". If this test is really that important I think we need to explain why.

daibach:
I don't *yet* know the reason this test is there. I'm going to find out, and see if we can update this to be more helpful.

For now though, make it my name.

jackscotti:
Yeah, I see your point, this should not happen. Updating the name for now, maybe @daibach has some context on this?

jackscotti:
👍 thank you both

PR #196, opened 2016-07-20
Fix gov_uk.feature

gpeng:
The recent features added to test the 'downcase all caps urls' functionality added in https://github.com/alphagov/govuk-puppet/pull/4573 were broken. This commit updates adds some steps to ensure the correct host is added and that the location is correctly checked.


rjw1:
looks fine to me.

PR #197, opened 2016-08-01
[DO NOT MERGE] Add test for query parameters on the mirrors

alexmuller:
We're unable to serve dynamic content from the static mirrors so we force every request with a query string to return a 503 error.


deanwilson:
Merge away

alexmuller:
Thanks, just waiting for the actual change to go in so that this will pass and then it can be merged.

deanwilson:
The changes to the mirror have been paused for now so this code should not be merged.

PR #198, opened 2016-08-02
Make Local Transaction test less brittle

sihugh:
This test uses a specific name for London Borough of Camden / Camden Borough
Council to verify the test.  We get council names from a variety of sources
and we want to somewhat insulate the test from the fact that these names change
from time to time.  

We are in the middle of switching from Publisher to Local Links Manager as the data source for Local Transaction links, and the data in each is slightly different due to the original sources disagreeing.


brenetic:
👍

PR #203, opened 2016-11-01
Fix the Trade Tariff search

kevindew:
It has changed from using a GET request to a POST request

We're not sure if this is supposed to be the case and this may need reverting if/when Trade Tariff is changed back to using GET.


andrewgarner:
👍 

This will be fine with either GET/POST as you are interacting with the user interface in the same way a human would.

PR #205, opened 2016-11-04
Disable licensing tests on integration

tijmenb:
These tests fail on integration.

Quote @timblair:

> So that's trying to download the relevant PDF form from Apto (the 3rd party that manages the PDFs for us), and it's failing, probably because the database in integration doesn't have the correct details in it.

It's on Tim's todo list now. In the meantime, this commit disables the test temporarily.

### DIFF
--- a/features/licensing.feature
+++ b/features/licensing.feature
@@ -1,6 +1,6 @@
 Feature: Licensing
 
-  @normal
+  @normal @notintegration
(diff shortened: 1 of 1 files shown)


alexmuller:
I'd rather we didn't do this kind of thing but I understand why we have in this case.

PR #211, opened 2016-11-30
Use cucumber profiles for environment specific behaviour

cbaines:


cbaines:
Now tested this on integration. There were some weird errors when puppet hadn't been deployed, and so the profile was not being specified:
```
No such file or directory @ rb_sysopen - . Please create a  directory to get started. (Errno::ENOENT)
```

Apart from that, when puppet finished deploying, the fix worked, and the tests for integration were disabled.

PR #213, opened 2016-12-07
Fix place postcode request test

davidbasalla:
Frontend now uses a GET request when the user searches for a postcode (not a POST request anymore). This commit makes the smokey tests pass again.


whoojemaflip:
👍

PR #215, opened 2016-12-09
Update find my nearest test to make html request

davidbasalla:
This was previously making a JSON request and testing some little used functionality in Frontend, which we did not test for in the app itself. We checked in Kibana and apart from Smokey no-one else appears to be posting to JSON place routes. We've chosen to remove this functionality as it simplifies the code in Frontend and do not want to be supporting legacy APIs just for the sake of it.

### DIFF
--- a/features/frontend.feature
+++ b/features/frontend.feature
@@ -58,15 +58,11 @@ Feature: Frontend
 
   @normal
   Scenario: check find my nearest returns results
-    When I try to post to "/ukonline-centre-internet-access-computer-training.json" with "postcode=WC2B+6NH"
+    When I visit "/ukonline-centre-internet-access-computer-training"
+    And I should see "UK online centres"
+    When I try to post to "/ukonline-centre-internet-access-computer-training" with "postcode=WC2B+6NH"
     Then I should get a 200 status code
     And I should see "Holborn Library"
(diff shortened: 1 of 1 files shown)


brenetic:
💭

features/frontend.feature
```diff
@@ -58,15 +58,11 @@ Feature: Frontend
...
   Scenario: check find my nearest returns results
-    When I try to post to "/ukonline-centre-internet-access-computer-training.json" with "postcode=WC2B+6NH"
-    Then I should get a 200 status code
-    And I should see "Holborn Library"
-
-  @normal
-  Scenario: check find my nearest (places) load
     When I visit "/ukonline-centre-internet-access-computer-training"
-    Then I should get a 200 status code
     And I should see "UK online centres"
```
whoojemaflip: Should this be a `Then` step?

features/frontend.feature
```diff
@@ -58,15 +58,11 @@ Feature: Frontend
...
+    When I try to post to "/ukonline-centre-internet-access-computer-training" with "postcode=WC2B+6NH"
```
whoojemaflip: "Do. Or do not. There is no try."

PR #218, opened 2016-12-19
Update Smokey tests for Design Principles

36degrees:
`design-principles` is no longer involved in serving the service manual – it’s entirely served by `service-manual-frontend`. Move the service manual homepage test from the design principles feature to the service manual feature.

Additionally, the search for the service manual is now handled by the global `/search` in combination with `filter_manual` so this does not need to be tested separately.

### DIFF
--- a/features/design_principles.feature
+++ b/features/design_principles.feature
@@ -11,19 +11,6 @@ Feature: Design Principles
     And I should see "Start with user needs"
 
   @normal
-  Scenario: check Service Manual
-    When I visit "/service-manual"
-    Then I should get a 200 status code
-    And I should see "Government Service Design Manual"
-
-  @normal
-  Scenario: check Service Manual search
-    When I visit "/service-manual/search?q=alpha"
-    Then I should get a 200 status code
-    And I should see "alpha"
-    And I should see some search results
-
-  @normal
   Scenario: check Transformation dashboard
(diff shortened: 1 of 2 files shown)


suzannehamilton:
Do we still need to do something about the 404s? It sounds like we should makes sure the old URLs still return results (presumably with a redirect), or return a 410.

36degrees:
We're working on redirecting requests to `/service-manual/search?q=whatever` to `/search?q=whatever&filter_manual=%2Fservice-manual` in service manual frontend.

features/design_principles.feature
```diff
@@ -11,19 +11,6 @@ Feature: Design Principles
     And I should see "Start with user needs"
 
   @normal
-  Scenario: check Service Manual
-    When I visit "/service-manual"
-    Then I should get a 200 status code
-    And I should see "Government Service Design Manual"
-
-  @normal
-  Scenario: check Service Manual search
```
36degrees: It seems redundant to me – search is independently tested and we're just using that with a filter_manual parameter.

tijmenb: 👍 agree

mcmillan:
@suzannehamilton There's a PR outstanding (alphagov/service-manual-publisher#89) which will redirect old searches to `/search`.

PR #221, opened 2017-01-12
Make smokey fail on javascript errors

tijmenb:
This enables turning on javascript for all tests.

This is what it looks like when there is an error:

screen shot 2017-01-12 at 17 07 18

This is follow-up from an incident that occured a few weeks week, where a rogue jQuery upgrade broke a number of pages on GOV.UK.

Trello: https://trello.com/c/WIunrfiT

PR #222, opened 2017-01-13
Smoke test for A/B testing

suzannehamilton:
Add feature tests for a help page which will display the state of an example A/B test (and potentially all A/B tests operating on gov.uk).

This test relies on A/B testing cookies being handled correctly by the server. This requires some CDN Varnish config (alphagov/govuk-cdn-config#13) which will only work on Production (and possibly Staging).

Since this doesn't work in dev, and since the real page doesn't exist yet, I've developed the tests against some temporary changes to collections (see the [ab-smokey-dev branch](https://github.com/alphagov/collections/tree/ab-smokey-dev)).

Both feature tests make multiple requests for statistical sampling. I've set them to make 20 requests each, which takes about 10 seconds per test. I think 20 requests should be high enough - it gives us  10^6 combinations of A/B variants, which means there's a 1 in 500,000 chance of a test giving the wrong result.

https://trello.com/c/MqHpmZYB/309-create-smokey-scenario-for-a-b-testing

### DIFF
--- a/features/step_definitions/ab_testing_steps.rb
+++ b/features/step_definitions/ab_testing_steps.rb
@@ -1,0 +1,58 @@
+Given(/^there is an AB test setup$/) do
+  # Empty step.
+  # We assume that there is always an A/B test set up on the example A/B test
+  # page. If this is not true, the A/B smoke tests will fail, so we should fix
+  # or delete them as appropriate.
+end
+
+Given(/^I do not have any AB testing cookies set$/) do
+  # Empty step.
+  # By default, no cookies are set.
+end
+
+Then(/^we have shown them all versions of the AB test$/) do
+  buckets = @responses.map { |r| ab_bucket(r.body) }.to_set
+  buckets.should == (Set.new ["A", "B"])
+end
+
+Then(/^I am assigned to a test bucket$/) do
+  ab_cookies = @response.headers[:set_cookie]
+    .select { |h| h.start_with? "ABTest-Example=" }
+  assert_equal 1, ab_cookies.length,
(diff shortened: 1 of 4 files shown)


suzannehamilton:
Just realised the seal was not picking up this PR because of the [DO NOT MERGE] marker. @tijmenb and/or @carvil - would you like to take a look?

features/step_definitions/ab_testing_steps.rb
```diff
@@ -0,0 +1,58 @@
...
+  assert bucket == "A" || bucket == "B",
+    "Expected A/B bucket to be 'A' or 'B', but got '#{bucket}'"
+
+  # Store bucket so that subsequent responses can be compared to the original
+  @original_bucket = bucket
+end
+
+Then(/^the bucket is reported to Google Analytics$/) do
+  # FIXME: Implement once we know how we're reporting to GA. Is it sufficient to
+  # check the meta tag which is inspected by GA to report the page view?
```
carvil: What would be our options here if we wanted to check that GA is actually called? Can we access the array of notifications we push to GA with phantomjs and assert that ours is there?

If we keep the markup but somehow change something that means GA stops looking into it, we could have a false positive.

tijmenb: In https://github.com/alphagov/smokey/pull/221 I'm working on adding javascript testing. One way to do it then would be to turn on analytics debugging and inspect `console.log` messages.

carvil: Cool, that would be good 👍

PR #227, opened 2017-01-25
Remove duplicate tag in runner script.

danielroseman:
The `@pending` tag is specified in the yaml configuration and also in
the command-line options that are passed from tests_json_output.sh. It
seems that specifying a tag twice causes Cucumber to skip all tests.
Removing the duplicate fixes the problem.


leenagupte:
👍 LGTM

PR #228, opened 2017-01-25
Revert switch to capybara-webkit

tijmenb:
This reverts https://github.com/alphagov/smokey/pull/222 and https://github.com/alphagov/smokey/pull/224.

The required packages still haven't been installed on the appropriate machines because govuk-puppet can't be deployed. This means production shows a lot of alerts and we won't get warnings for actual failures. Additionally these new tests seem to be failing for no reason on integration https://deploy.integration.publishing.service.gov.uk/job/Smokey/13555/console.

Tested this: https://deploy.integration.publishing.service.gov.uk/job/Smokey/13557/console


leenagupte:
👍

PR #230, opened 2017-02-01
Enable JavaScript testing with Poltergeist

alecgibson:
This Pull Request enables JavaScript testing in Smokey with the Poltergeist web driver.

It does this in two steps:
1. Unrevert @tijmenb 's original PR achieving this with Capybara-Webkit
2. Switch the web driver over to Poltergeist.

The change to Poltergeist has been made in the hopes of improved stability. For example, with Capybara-Webkit the Scenario `Signon cookies are marked as secure and HttpOnly` would fail intermittently. This was reproducible on a Development environment running against the Integration server.

The (immediate) cause was that Capybara-Webkit would not populate the `Capybara::Session` object correctly (for example, the body, headers and status code of the response would be missing, but the title would be present, and a screenshot could be output displaying the expected page). I could not figure out anything deeper than this cause (which would presumably involve rooting around in Capybara-Webkit itself).

In an attempt to improve stability, I have changed the web driver to Poltergeist. Running under the same conditions, I have not been able to reproduce the previously failing intermittent error.

Trello: https://trello.com/c/WIunrfiT/322-investigate-turning-on-javascript-in-smokey


features/whitehall.feature
```diff
@@ -132,5 +132,4 @@ Feature: Whitehall
     And I force a varnish cache miss
     And I am benchmarking
     When I visit "/banknote"
-    # Redirects are transparently followed, so the end status code is 200
```
tijmenb: This test shouldn't be modified, for "reasons".

alecgibson: I know, I've sent @davidillsley a Slack (and CC here for good measure)

features/step_definitions/smokey_steps.rb
```diff
@@ -203,3 +208,7 @@ def random_path_selection(opts={})
   anchor_tags = opts[:anchor_tags] || []
   anchor_tags.map { |anchor| anchor.attributes["href"].value }.sample(size)
 end
+
+When /^I inject a JavaScript error on the page, Smokey raises an exception$/ do
+  expect { page.driver.execute_script('1.error') }.to raise_error
+end
```
tijmenb: No trailing line here (we should add govuk-lint to this repo).

features/javascript_errors.feature
```diff
@@ -0,0 +1,5 @@
+Feature: Smokey Javascript Error detection
+
+  @normal
+  Scenario: Smokey detects JS errors
+    When I inject a JavaScript error on the page, Smokey raises an exception
```
tijmenb: This is clever!

features/whitehall.feature
tijmenb: I already talked to him a couple of weeks ago. Let's not change it.

alecgibson: I think I have to change it. For some reason, Poltergeist will successfully redirect pages, but won't update the status code from `301` to `200`. As pointed out [here|https://github.com/teamcapybara/capybara/issues/1149], integration tests like this should be using a website how a user would use a website - and that doesn't include looking at status codes. (Arguably better practice still than checking the path is checking for a known element on the page to ensure it's actually been rendered).

Why can't we change it?

jenkins.sh
```diff
@@ -18,5 +18,5 @@ if [ ! -f .ruby-version ]; then
   export PATH=$(printf $PATH | awk 'BEGIN { RS=":"; ORS=":" } !/rbenv/' | sed 's/:$//')
 fi
 
-bundle install --path "${HOME}/bundles/${JOB_NAME}" --deployment --quiet
-RESTCLIENT_LOG="log/smokey-rest-client.log" govuk_setenv default bundle exec rake $MYTASK
+bundle install --path "${HOME}/bundles/${JOB_NAME}" --deployment
+RESTCLIENT_LOG="log/smokey-rest-client.log" govuk_setenv default xvfb-run -a bundle exec rake $MYTASK
```
tijmenb: `xvfb-run -a` is no longer  necessary. It was added to make the tests run on webkit.

alecgibson:
@tijmenb is this good to go?

features/javascript_errors.feature
fofr: Nicely done!

features/step_definitions/efg_steps.rb
```diff
@@ -1,24 +1,23 @@
 Given /^I am testing in an EFG context$/ do
-  # Not sure if there is a better way to do this?
-  page.driver.browser.agent.add_auth(efg_base_url, ENV['AUTH_USERNAME'], ENV['AUTH_PASSWORD'])
+  # TODO: remove me
```
Davidslv: was this forgotten?

alecgibson: Not sure - @tijmenb ?

PR #232, opened 2017-02-07
Temporary JS error ignore

alecgibson:
Add a temporary fix for a test getting JS errors, because some JavaScript somewhere is using `Function.prototype.bind`, which we technically do not support. I will try to hunt down this usage and remove it, so we can remove this workaround.


alecgibson:
Superseded by: https://github.com/alphagov/smokey/pull/234

PR #234, opened 2017-02-07
Restore Poltergeist and fix Smokey tests

alecgibson:
This PR puts back the update to use JS via Poltergeist, also fixing tests that errored beyond Integration.


carvil:
This now works as expected 👍

PR #231, opened 2017-02-06
[DO NOT MERGE] Add test for A/B test Google Analytics tracking

alecgibson:
This test checks that we are correctly sending a page view tracking event to Google Analytics when A/B testing.

Note that this card is dependent upon:
- [x] https://github.com/alphagov/smokey/pull/230
- [ ] https://github.com/alphagov/static/pull/888

Trello: https://trello.com/c/g9OyC7Pn/310-tell-google-analytics-about-a-b-tests

PR #236, opened 2017-02-07
Add test for Google Analytics AB tracking

alecgibson:
This test checks that we are correctly sending a page view tracking event to Google Analytics when A/B testing.

Note that this card is dependent upon:
- [x] https://github.com/alphagov/smokey/pull/230
- [x] https://github.com/alphagov/static/pull/888

Follow-up from: https://github.com/alphagov/smokey/pull/231 (the branch it was based on was deleted)

Trello: https://trello.com/c/g9OyC7Pn/310-tell-google-analytics-about-a-b-tests


tijmenb:
https://github.com/alphagov/static/pull/888 has just been deployed.

PR #238, opened 2017-02-08
[Do not merge] Fix teardown for ignoring JS errors

alecgibson:
A previous step was introduced to ignore JavaScript errors, but this step did not correctly teardown (so _all_ subsequent tests also ignored JavaScript errors).

This change correctly tears down the error ignoring through the use of a tag, and adds a test to check this teardown.


alecgibson:
Merging is blocked on getting Poltergeist back into Smokey (it was reverted here: https://github.com/alphagov/smokey/pull/237)

alecgibson:
Superseded by https://github.com/alphagov/smokey/pull/239

PR #239, opened 2017-02-08
Suppress PhantomJS logging

alecgibson:
Smokey was previously updated to include Poltergeist for Javascript support, but this was reverted because the logging from PhantomJS was incompatible with our own monitoring script (https://github.com/alphagov/smokey/pull/237)

This puts Poltergeist back, as well as a fix for JavaScript error ignoring, and dumps all PhantomJS logging into `logs/phantomjs.log`.

Note that we should raise separate tickets for dealing with the warnings that arise in this log.


tijmenb:
Confirmed on production:

https://deploy.publishing.service.gov.uk/job/Smokey/4471

And staging:

https://deploy.staging.publishing.service.gov.uk/job/Smokey/5161/

PR #241, opened 2017-02-10
Stop benchmarking in smoke tests

tijmenb:
This is controversial - I'm totally fine with this not being merged. 

I believe we should not benchmark pages in smokey.

Firstly, because the tests are brittle and often fail. This erodes confidence in the tests, which leads people to ignore them. When something real breaks, we might not know. Perhaps intermittently failing tests are worse than no tests.

Secondly, smokey is not a good tool to verify the performance of an application. Slowdowns can have many causes. When smokey thinks something is too slow, we learn nothing about the cause.

Lastly, developers can't reasonably act on these alerts. We've seen dozens of smokey fails over the last months, since we started broadcasting failures on Slack. Nothing has happened to fix the underlying cause, probably because there's not actually a real problem here.

### DIFF
--- a/features/bouncer.feature
+++ b/features/bouncer.feature
@@ -3,8 +3,6 @@ Feature: Bouncer
   @high
   Scenario: Bouncer application is up
     Given I am testing "bouncer"
-    And I am benchmarking
     When I request "http://www.attorneygeneral.gov.uk" from Bouncer directly
     Then I should get a 301 status code
     And I should get a "Location" header of "https://www.gov.uk/government/organisations/attorney-generals-office"
-    And the elapsed time should be less than 2 seconds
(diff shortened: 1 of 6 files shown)


boffbowsh:
I agree with this in principle - however I don't think we currently have another place to expose and record application slowness. Smokey is not the correct place for it, but I think until we have some sort of dashboard that says that a particular application is/has become slow, we should leave it in Smokey. 
While these thresholds are arbitrary, they're a good indicator right now of something being wrong. Most of the time, the failures will disappear on next run if there's just an odd spike. If the alert doesn't disappear, 2ndline should and will investigate.
Also happy to hear arguments against.

tijmenb:
Yep, I agree with all that. I guess the problem is that 2nd line doesn't investigate smokey benchmark failures, because odd spikes sometimes appear & disappear.

Observing the behaviour of smokey over the last months (since slack), I would say that the benchmarking tests are _not_ a good indicator of something being wrong. They're consistently giving us the wrong message.

Appeal to authority:

> The trouble with non-deterministic tests is that when they go red, you have no idea whether its due to a bug, or just part of the non-deterministic behavior. Usually with these tests a non-deterministic failure is relatively common, so you end up shrugging your shoulders when these tests go red. Once you start ignoring a regression test failure, then that test is useless and you might as well throw it away.[3]
>
> Indeed you really ought to throw a non-deterministic test away, since if you don't it has an infectious quality. If you have a suite of 100 tests with 10 non-deterministic tests in them, than that suite will often fail. Initially people will look at the failure report and notice that the failures are in non-deterministic tests, but soon they'll lose the discipline to do that. Once that discipline is lost, then a failure in the healthy deterministic tests will get ignored too. At that point you've lost the whole game and might as well get rid of all the tests.


https://martinfowler.com/articles/nonDeterminism.html

boffbowsh:
To me, the intermittent failures are blink-and-you-miss-it. If it's still around 5 minutes later then you should investigate.

The icinga Smokey alerts are probably fine, and the Jenkins ones are probably not. So maybe just ignore them in Jenkins?

tijmenb:
Aha, that's where the different outlook comes from. That's a nice idea, I'll investigate.

PR #242, opened 2017-02-13
Use the correct A/B test name in Analytics test

tijmenb:
@alecgibson @suzannehamilton right?


suzannehamilton:
Pretty sure it needs to be 'Example' because everything needs to match the cookie name.

I think this means that frontend is setting the name incorrectly now we've dropped the `.capitalize` from the A/B gem.

suzannehamilton:
I've just submitted a PR to fix frontend: alphagov/frontend#1126

PR #248, opened 2017-02-28
Revert "Update Gems"

davidbasalla:
Reverts alphagov/smokey#247

The gem bump for rest-client gives the following errors. We should revert this until we get time to fix this.

```
2017-02-28 15:02:11: Scenario: check licences load (features/frontend.feature/@normal)
2017-02-28 15:02:11: Step: [PASS] Given I am testing through the full stack
2017-02-28 15:02:11: Step: [PASS] And I force a varnish cache miss
2017-02-28 15:02:11: Step: [PASS] When I visit "/busking-licence"
2017-02-28 15:02:11: Step: [PASS] Then I should see "Busking licence"
2017-02-28 15:02:11: Step: [PASS] And I should see an input field for postcode
2017-02-28 15:02:11: Step: [FAIL] When I try to post to "/busking-licence" with "postcode=E20+2ST"
2017-02-28 15:02:11: Error: wrong number of arguments (given 2, expected 0) (ArgumentError)
2017-02-28 15:02:11: Step: [SKIP] Then I should get a 200 status code
2017-02-28 15:02:11: Step: [SKIP] And I should see "Busking licence"
```

PR #249, opened 2017-02-28
Upgrade nokogiri and rest-client to appease Synk

issyl0:
- Synk reports that rest-client 1.8.0 is not vulnerable. This RestClient
  bump is a minor version bump, not a major one, and so thankfully does not
  cause our tests to break.

This is an alternative to #248 which does not reintroduce the vulnerability.


sihugh:
:1st_place_medal:

PR #251, opened 2017-03-08
Fix collections test

alecgibson:
This fixes a failing Smokey test for collections, which was not correctly setting up an A/B test

### DIFF
--- a/features/step_definitions/ab_testing_steps.rb
+++ b/features/step_definitions/ab_testing_steps.rb
@@ -13,7 +13,11 @@ Given(/^I do not have any AB testing cookies set$/) do
 end
 
 Given(/^I am in the "(A|B)" group for "(.*)" AB testing$/) do |test_group, ab_test|
+  # Set BOTH the cookie and the header, because:
+  # - On development/int/staging, cookies don't work so use the header directly
+  # - On production, the header is ignored and overwritten by the CDN, but honours the cookie
   page.driver.set_cookie("ABTest-#{ab_test}", test_group)
+  page.driver.add_headers("GOVUK-ABTest-#{ab_test}" => test_group)
 end
 
 Then(/^we have shown them all versions of the AB test$/) do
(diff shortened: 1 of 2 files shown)


features/step_definitions/ab_testing_steps.rb
```diff
@@ -13,7 +13,7 @@
 end
 
 Given(/^I am in the "(A|B)" group for "(.*)" AB testing$/) do |test_group, ab_test|
-  page.driver.set_cookie("ABTest-#{ab_test}", test_group)
+  page.driver.add_headers("GOVUK-ABTest-#{ab_test}" => test_group)
```
suzannehamilton: I suspect the old version only worked on production, and the new version will only work on dev/integration/staging.

On production, the header will be overridden by the CDN because it has no cookie set.

I think we should set both and add a comment explaining why. It's probably safe to use both as long as the ab_testing.feature tests don't call this step - those tests really should be testing the whole stack, rather than setting intermediate A/B testing headers.

alecgibson: Done.

features/step_definitions/ab_testing_steps.rb
```diff
@@ -13,7 +13,11 @@
 end
 
 Given(/^I am in the "(A|B)" group for "(.*)" AB testing$/) do |test_group, ab_test|
+  # Set BOTH the cookie and the header, because:
+  # - On development/int/staging, cookies don't work so use the header directly
+  # - On production, the header is stripped, but cookies work
```
suzannehamilton: Strictly speaking it's not stripped, it's just overwritten either by the existing cookie value or a random new value. (I know that's pedantic but the CDN config is confusing enough as it is, so I think it's worth making sure we describe it correctly.)

alecgibson: How about now?

PR #252, opened 2017-03-09
Test that we place users in B variant on new nav

carvil:
Regardless of a user having a cookie or not, when they go to the new
education pages in collections, they should be assigned to the B
variant.

These tests will verify that the VCL configuration is working as
expected.

This depends on https://github.com/alphagov/govuk-cdn-config/pull/35
being live.

### DIFF
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -18,7 +18,6 @@ Feature: Collections
     Then I am in the "A" variant of the education navigation test
     And I stay on bucket "A" of the education navigation test when I keep visiting "/education-maintenance-allowance-ema"
 
-  @notstaging @notproduction
   Scenario: Using the accordion page
     Given I am in the "B" group for "EducationNavigation" AB testing
     When I visit "/education/special-educational-needs-and-disability-send-and-high-needs"
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -1,5 +1,23 @@
 Feature: Collections
 
+  @notintegration @notstaging
+  Scenario: Automatic opt-in to the B variant from navigation pages without any
+    previous variant
+    Given I do not have any AB testing cookies set
+    And I am testing through the full stack
+    When I visit "/education"
+    Then I am in the "B" variant of the education navigation test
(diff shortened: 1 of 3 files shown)


features/collections.feature
```diff
@@ -1,5 +1,30 @@
...
+    When I visit "/education/running-and-managing-a-school"
+    Then I am in the "B" variant of the education navigation test
+    And I stay on the same bucket when I keep visiting "/education/running-and-managing-a-school"
+
+  Scenario: Don't automatically opt-in to B variant on content pages
+    that start with "education"
+    Given I am in the "A" group for "EducationNavigation" AB testing
+    When I visit "/education-maintenance-allowance-ema"
+    Then I am in the "A" variant of the education navigation test
+    And I stay on the same bucket when I keep visiting "/education"
```
suzannehamilton: Should this path match line 24? Otherwise I'm guessing this step will fail because visiting /education will put you in B.

carvil: Thanks for spotting it, I've pushed a fix.

carvil:
@tijmenb / @suzannehamilton I've made some changes, including enabling the accordion test in all environments. The new tests work in production (and staging when pointing to `www`).

Thought I should give you a nudge as you've approved this before.

features/step_definitions/collections_steps.rb
```diff
@@ -21,3 +21,31 @@
...
+  assert(
+    ab_cookie.expires,
+    "Expected the cookie to have an expiry date"
+  )
+end
+
+Then(/^I stay on bucket "(.*?)" when I keep visiting "(.*?)"$/) do |variant, path|
+  20.times do
+    response = visit path
+    education_ab_cookie = page.driver.cookies['ABTest-EducationNavigation']
```
suzannehamilton: Really minor point: could the step name mention education navigation to make it clearer that the step checks for this specific cookie?

carvil: Yup, I will change the step name.

PR #257, opened 2017-04-20
Fix Local Links Manager failures

issyl0:
- Since we changed the style of the homepage to be broken links
  instead of a list of LAs, this has been failing. Interestingly,
  I think this has only been failing in Integration, yet in the
  version of LLM we have in Production I can only see lowercase
  "local authorities" which does not fit with what the test is
  asking for (capitalised). Hmm.


issyl0:
We could have made this test check for "Broken links", but we're still changing the display of that page, so one of the other (more consistent) homepage tab names - "Councils" or "Services" - was deemed a better option.

PR #258, opened 2017-04-20
Add Cucumber scenario for manuals-publisher

chrisroos:
This follows the pattern of the other publishing tests.

One of the motivations for adding a Smoke test was to catch the sort of timeout
problem we saw recently where the /manuals index page was taking too
long to load. Unfortunately, I don't think this test would've caught
that problem because the test signon user doesn't have any manuals to
load! Irrespective, I still think it's useful to have this sort of test
to ensure we get an early warning if we've really broken something.

The test signon user has the manuals-publisher permission. You can run
this test locally by using the signon credentials stored in the Smokey
job configuration in the integration Jenkins instance.

I've set the scenario as `@high` priority which means (if I understand [the
README][1] correctly) that it'll generate a pager alert if it fails. This
is the same priority as the other scenarios in this feature although
it's not obvious to me that it's definitely the right thing to do.

[1]:
https://github.com/alphagov/smokey/blob/master/README.md#prioritising-scenarios

### DIFF
--- a/features/mainstream_publishing_tools.feature
+++ b/features/mainstream_publishing_tools.feature
@@ -34,3 +34,13 @@ Feature: Mainstream Publishing Tools
     Then I should see "Local Links Manager"
       And I should see "Sign out"
       And I should see "Local Authorities"
+
+  @high
+  Scenario: Can log in to manuals-publisher
+    When I go to the "manuals-publisher" landing page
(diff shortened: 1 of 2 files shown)


chrisroos:
Thanks @floehopper. I'm going to get this merged.

Issue #256, opened 2017-04-11
Smokey hits possibly showing up in analytics


fofr:
Since we turned on Javascript we've seen a spike in traffic to our admin analytics profile. Smokey is the most likely cause. This extra traffic makes observing real user behaviour a little difficult.

I believe the best thing to do would be to filter out the traffic in GA, but I'm raising the issue here to provide a place to track it.

![screen_shot_2017-04-11_at_15 41 52](https://cloud.githubusercontent.com/assets/319055/24915328/c6450686-1ece-11e7-898b-c5465e92127d.png)

tijmenb:
Should have been prevented by https://github.com/alphagov/smokey/pull/245. @carvil any ideas?

carvil:
@tijmenb / @fofr apologies for the delay in getting back to you.

Could you please let me know what sort of requests we are getting in GA? We blacklisted the analytics' URL, so I'm interested in seeing which tests are doing this.

As an example, when running the collections' feature locally (pointing to staging) with the blacklist URL, we can see the following hosts in the network traffic:

```
assets-origin.staging.publishing.service.gov.uk
www.staging.publishing.service.gov.uk
```

However, if I delete the blacklist URLs, this is what I see:

```
assets-origin.staging.publishing.service.gov.uk
stats.g.doubleclick.net
www.google-analytics.com
www.google.co.uk
www.google.com
www.staging.publishing.service.gov.uk
```

The only test that bypasses the blacklist URL is one A/B testing test, so I'm wondering if that's the one causing the issues: 

https://github.com/alphagov/smokey/search?utf8=%E2%9C%93&q=withanalitics&type=

fofr:
They were mostly requests to Sign on, on `/users/sign_in` and `/`, across a variety of platforms (integration, prod, staging).

carvil:
@fofr the existing configuration does work on individual test runs, but not when the whole suite runs. The PR #260 is an attempt at fixing this by adding the blacklist URL list before each relevant scenario, and raising if it finds the google analytics' host in the network traffic after the scenarios run.

carvil:
@fofr the PR has been merged and I ran it on all environments successfully. We should check Analytics again early next week to verify if things have gone back to normal.

fofr:
Looking good. 👍 💯 📉 

![screen shot 2017-04-24 at 10 55 06](https://cloud.githubusercontent.com/assets/319055/25331889/839e5eb4-28dc-11e7-966b-e3f34da513c0.png)

PR #260, opened 2017-04-21
Disable analytics before each scenario

carvil:
This has the exception of 1 test case where we need to send an analytics
event, but it should disable Google Analytics across all other tests.

I've also added a User-Agent to poltergeist so that we can recognise
those requests, should we need to.

Finally, we will raise an error if the Google Analytics host is somehow
in the list of hosts from the driver's network traffic list. We don't
want for this to happen and there is a cucumber tag to enable it, in
case someone needs it (e.g. used for one A/B test).

This should fix issue #256

cc/ @tijmenb and @fofr


fofr:
Top sleuthing :+1:

PR #261, opened 2017-04-25
Remove draft test for contacts-frontend

theleebriggs:
# Description

`contacts-frontend` has been replaced with `government-frontend` and is now being retired. This PR removes tests specific to `contacts-frontend`.

### DIFF
--- a/features/draft_environment.feature
+++ b/features/draft_environment.feature
@@ -32,9 +32,3 @@ Feature: Draft environment
     Then I should see "Content design"
     And I should see "Search this manual"
     And the page should contain the draft watermark
-
-  @draft
-  Scenario: visiting a page served by contacts-frontend
-    When I try to login as a user
-    When I attempt to visit "government/organisations/hm-revenue-customs/contact/child-benefit"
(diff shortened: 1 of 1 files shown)


features/draft_environment.feature
```diff
@@ -32,9 +32,3 @@ Feature: Draft environment
     Then I should see "Content design"
     And I should see "Search this manual"
     And the page should contain the draft watermark
-
-  @draft
-  Scenario: visiting a page served by contacts-frontend
```
fofr: Presumably this is a test to check that contacts-frontend is running, rather than a test that contacts are successfully rendering. If it's the latter then the scenario could be renamed.

theleebriggs: I read the tests in that file to be testing rendering apps via the draft stack - there is already a `government-frontend` test in there. There are other features for testing general rendering:

* https://github.com/alphagov/smokey/blob/master/features/contacts.feature
* https://github.com/alphagov/smokey/blob/master/features/government_frontend.feature

PR #262, opened 2017-05-09
Add test for related link overrides for A/B test

carvil:
We have a bunch of overrides for related links on the new
Education Navigation A/B test in place across different apps. This is
hard to test in the apps as we mock requests/responses, and we've seen
it breaking when we moved the rendering app of some of those pages.

This test should be deleted once we have a solution for related links in
the new navigation.

Trello: https://trello.com/c/U6JKZN6s/115-make-sure-we-override-the-related-links-on-some-mainstream-content-that-was-recently-moved-to-government-frontend

cc/ @fofr

### DIFF
--- a/features/step_definitions/ab_testing_steps.rb
+++ b/features/step_definitions/ab_testing_steps.rb
@@ -70,6 +70,38 @@ Then(/^I stay on the same bucket when I keep visiting "(.*?)"$/) do |path|
   end
 end
 
+Then(/^I can see the old related links sidebar$/) do
+  refute(
+    page.has_selector?('.govuk-taxonomy-sidebar'),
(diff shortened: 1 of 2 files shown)


features/step_definitions/ab_testing_steps.rb
```diff
@@ -70,6 +70,18 @@
   end
 end
 
+Then(/^I can see the old related links sidebar$/) do
+  assert(
```
fofr: Would `refute` without the `== false` be better here?

carvil: Yes, I will make that change.

carvil:
When https://github.com/alphagov/government-frontend/pull/355 is merged this will be green in production.

PR #273, opened 2017-06-16
Add tests for the public API

rubenarakelyan:
This commit adds some tests to ensure the public API is available and returning JSON. It tests the search API, content API and the various whitehall APIs. It does not test whether the data returned by the APIs is correct.

It also moves the step definition for testing for JSON responses to a more generic file since it is now used in mutiple tests.

### DIFF
--- a/features/publicapi.feature
+++ b/features/publicapi.feature
@@ -1,0 +1,38 @@
+Feature: Public API
+
+    Background:
+      Given I am testing through the full stack
+
+    @normal
+    Scenario: Check the search API returns data
+      When I request "/api/search.json"
+      Then I should get a 200 status code
+      And JSON is returned
+
+    @normal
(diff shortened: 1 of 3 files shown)


features/publicapi.feature
```diff
@@ -0,0 +1,38 @@
...
+      Given I am testing through the full stack
+
+    @normal
+    Scenario: Check the search API returns data
+      When I request "/api/search.json"
+      Then I should get a 200 status code
+      And JSON is returned
+
+    @normal
+    Scenario: Check the content API returns data
```
tijmenb: Maybe "content store"?

rubenarakelyan: Naming is hard - https://github.com/alphagov/smokey/pull/274

tijmenb: ❤️

PR #276, opened 2017-07-20
Switch the case study used for the draft environment tests

cbaines:
To one that does not have JavaScript from YouTube on the page, as this
errors when tested by Smokey.


cbaines:
Thanks for reviewing @carvil :)

PR #280, opened 2017-08-02
Add youtube.com to Poltergeist's blacklist of URLs

chrisroos:
This fixes the intermittent error we've been seeing in the "check
Transformation dashboard" scenario of the "Design Principles" feature;
as reported in issue #279.

I ran the problematic scenario 10 times before this fix and 10 times
after:

    $ bundle exec cucumber features/design_principles.feature:14

I saw 6 failures and 4 successes before the fix and 10 successes
afterwards so I'm fairly confident this will solve the intermittent
failures we've been seeing.

Blacklisting youtube.com doesn't appear to negatively affect any of the
other scenarios.

PR #281, opened 2017-08-02
Simplify URL blacklisting configuration

chrisroos:
This is a precursor to PR #280.

I want to add 'www.youtube.com' to the URL blacklist but doing so made it apparent that the current implementation was quite closely tied to blacklisting 'www.google-analytics.com'. I think/hope this change makes the intention of the URL blacklisting a bit clearer.

This builds upon both PR #245 and PR #260.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -20,7 +20,10 @@ phantomjs_logger = File.open("log/phantomjs.log", "a")
 BLACKLISTED_URLS = ['www.google-analytics.com']
 
 Capybara.register_driver :poltergeist do |app|
-  Capybara::Poltergeist::Driver.new(app, phantomjs_logger: phantomjs_logger)
+  options = {
+    phantomjs_logger: phantomjs_logger
+  }
+  Capybara::Poltergeist::Driver.new(app, options)
 end
 
 Capybara.default_driver = :poltergeist
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -21,6 +21,7 @@ BLACKLISTED_URLS = ['www.google-analytics.com']
 
 Capybara.register_driver :poltergeist do |app|
   options = {
+    debug: ENV['POLTERGEIST_DEBUG'] || false,
     phantomjs_logger: phantomjs_logger
   }
   Capybara::Poltergeist::Driver.new(app, options)
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -44,10 +44,6 @@ After('~@withanalytics') do
   end
 end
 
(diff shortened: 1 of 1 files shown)


chrisroos:
Hi @fofr, @carvil and @tijmenb. I've requested your reviews on this PR as you've been involved in the previous PRs related to this functionality. I hope that's OK!

carvil:
@chrisroos I'm no longer at GDS, but happy to give a bit of context. We were spamming GA with Smokey and this was raised by the performance analysts some time ago. We decided to black list that URL to avoid that, but enable for the only tests where we need the analytics request to go through. If that scenario is kept, then it should be fine.

features/support/env.rb
```diff
@@ -17,33 +17,22 @@
...
-BLACKLISTED_URLS = ['www.google-analytics.com']
+GOOGLE_ANALYTICS_URL = 'www.google-analytics.com'
+BLACKLISTED_URLS = [GOOGLE_ANALYTICS_URL]
 
 Capybara.register_driver :poltergeist do |app|
-  Capybara::Poltergeist::Driver.new(app, phantomjs_logger: phantomjs_logger)
+  options = {
+    debug: ENV['POLTERGEIST_DEBUG'] || false,
+    phantomjs_logger: phantomjs_logger,
+    url_blacklist: BLACKLISTED_URLS
```
fofr: Is this what we used to do here:
https://github.com/alphagov/smokey/pull/260/files
https://github.com/alphagov/smokey/pull/260/files#diff-57013262a6cdf1f0f6cb2c64aa762c68L26

Which caused this: https://github.com/alphagov/smokey/issues/256

chrisroos: It's not quite the same as the original implementation (from PR #245), although it does look quite similar. I'm quite confident that this branch retains the behaviour of the fixed version (PR #260) and so we shouldn't see a repeat of issue #256.

Having said all that, you've made me want to double check the behaviour I'm seeing so that I can be _completely_ confident that I haven't reintroduced the problem described in issue #256.

chrisroos: I'm now completely confident that I haven't reintroduced the problem described in issue #256. I've updated the commit titled "Instantiate Poltergeist driver with url_blacklist" to explain the change in more detail.

chrisroos:
Thanks for the background, @carvil :-)

features/support/env.rb
fofr: Good commit message. 💯

chrisroos:
Thanks @fofr. I'll merge this and get PR #280 rebased on top of it.

PR #283, opened 2017-08-03
Use app_domain instead of GOVUK_APP_DOMAIN in licensing.rb

chrisroos:
The `#app_domain` method provides a sensible default if
`GOVUK_APP_DOMAIN` isn't set. Using this means that we can run the
"Signing in to licensify-admin" scenario of the Licensing feature
without having to explicitly set `GOVUK_APP_DOMAIN`.


chrisroos:
Hi @alecgibson. I've requested your review as GitHub tells me you've recently edited this file. I hope that's OK!

chrisroos:
Thanks @alecgibson! I'll get this merged.

PR #286, opened 2017-08-10
Add test for Smart Answers country pickers

alecgibson:
Following an incident where the country pickers were showing no (or very
few) countries on Smart Answers, this change adds tests to check that
the dropdowns are populated with a sensible-looking number of elements.

This test is needed, because Smart Answers relies on getting countries
from Whitehall, so we should test this integration.

The tests check for the number of countries being greater than 200,
which should provide a sense check without requiring constant updating
of the list of countries every time Whitehall or Smart Answers changes.


alecgibson:
[(2) Add smokey tests to check for empty Whitehall WorldLocation.all (i.e country list)](https://trello.com/c/gUBh60AK/708-2-add-smokey-tests-to-check-for-empty-whitehall-worldlocationall-ie-country-list)

PR #291, opened 2017-09-05
Check what facets look like in finder pages

MatMoore:
This tests for issues like this:
https://docs.publishing.service.gov.uk/manual/incorrect-content-in-search-or-navigation.html#there-are-blank-options-on-finder-pages

The root cause in this case is rummager holding inconsistent data.  It knows
pages are tagged to people/organisations/topics but sometimes it's missing the
thing being tagged to.

In this case finder frontend just renders the thing with a blank title.

We could handle this better in finder frontend, but I think if the labels are ever
blank it should fail smokey, as this means the finder is not working properly.

Trello: https://trello.com/c/oTkICgvO/291-stop-blank-people-appearing-in-policy-finder

### DIFF
--- a/features/finder_frontend.feature
+++ b/features/finder_frontend.feature
@@ -1,5 +1,5 @@
 Feature: Finder Frontend
-  This are pages that have been migrated to a finder instead of index pages.
+  These are pages that let you search within a bunch of similar looking documents
 
   Background:
     Given I am testing through the full stack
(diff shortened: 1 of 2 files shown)


dwhenry:
Have we seen this test fail as expected when incorrect values are returned from ES?

Rosa-Fox:
Looks good to me. @dwhenry I checked with Mat and he said it was fine as was passing locally for him with incorrect values.

PR #293, opened 2017-09-07
Check that search results are not duplicated

MatMoore:
This only checks a couple of common search terms, but should catch some obvious bugs
affecting high volume content.

Note: This test currently fails due to a bug with search we are fixing. https://trello.com/c/jx2JMs9R/301-prevent-duplicate-search-results-for-smart-answers

### DIFF
--- a/features/search.feature
+++ b/features/search.feature
@@ -1,11 +1,28 @@
 Feature: Search
 
   @high
-  Scenario: check search results
+  Scenario: check search results for tax
     Given I am testing through the full stack
     And I force a varnish cache miss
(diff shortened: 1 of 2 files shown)


dwhenry:
Could potentially rewrite to be a scenario outline and remove some duplication: https://github.com/cucumber/cucumber/wiki/Scenario-Outlines

MatMoore:
@dwhenry Ah nice, I was wondering if there's something like that. I'm going to merge for now, and raise a follow up PR for this (this will let me concisely add a few more examples as well).

PR #294, opened 2017-09-08
Use scenario outline for checking search

MatMoore:
Use a range of common queries with varying query types
(acronyms, single words, multi word)

### DIFF
--- a/features/search.feature
+++ b/features/search.feature
@@ -1,28 +1,26 @@
 Feature: Search
 
   @high
-  Scenario: check search results for tax
+  Scenario Outline: check search results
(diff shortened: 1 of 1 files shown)


cbaines:
I remember having problems with using a scenario outline, but things may have changed, as it was a little while ago a7eb073ac62efaac408e2a9af989200b2a2af699

MatMoore:
☹️  I'm going to close this for now, since I'd rather investigate the nagios_check_cache.py issue before deploying, but I'm not going to get around to this straight away.

With more examples, it does take longer to run, but I feel like the current number is a good trade off.

PR #297, opened 2017-10-10
Add test for CSV preview

rubenarakelyan:
This commit adds a test for the whitehall CSV preview to ensure JavaScript is running without any errors on those pages.

Trello: https://trello.com/c/BoQt5GLG/195-add-smokey-test-for-csv-preview-incident-review-action

### DIFF
--- a/features/step_definitions/csv_steps.rb
+++ b/features/step_definitions/csv_steps.rb
@@ -1,0 +1,5 @@
+Then /^JavaScript should run without any errors$/ do
+  # If the `report-a-problem-toggle-wrapper` element exists then JavaScript
+  # is running on the page and there are no SRI or similar errors
+  expect(page).to have_selector('.report-a-problem-toggle-wrapper')
+end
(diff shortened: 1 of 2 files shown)


features/step_definitions/csv_steps.rb
```diff
@@ -0,0 +1,5 @@
+Then /^JavaScript should run without any errors$/ do
+  # If the `js-enabled` class has been set on the <body> element then
+  # JavaScript is running on the page and there are no SRI or similar errors
+  expect(page.find('body')[:class].split(" ")).to include("js-enabled")
```
kevindew: I'm not super knowledgeable about the issues behind this - so this comment might be hopelessly wrong - but I'm not sure how this test helps us with errors.

as I understand (from looking at source) that this line after the opening `<body>` element: `<script>document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');</script>` modifies the className and that will run regardless (and before) any external javascript files which have an integrity hash.

rubenarakelyan: My mistake. I've changed it to look for something that we add via an external JS file.

PR #298, opened 2017-10-12
Don't run mirrorer tests on AWS

issyl0:
- We don't have a VPN set up from AWS to the mirror boxes, so these
  Smokey tests are failing. Avoid this noise in our AWS environment by
  disabling these tests in that case.


issyl0:
Tested this on AWS and the alert went away.

issyl0:
Ahh, right. Yep. Thanks for the review!

PR #307, opened 2017-12-08
Allow setting of CUSTOM_DESTINATION_URI

surminus:
For the migration to AWS, we need to test the service domain works correctly before we do the switchover. To achieve this, we want to use Smokey as an indicator that stuff works like we expect.

Add an environment variable called `CUSTOM_DESTINATION_URI` that allows setting a custom domain. For our use, it will be something using the top level "integration.govuk.digital" domain, while setting the `GOVUK_WEBSITE_ROOT` and `EXPECTED_GOVUK_WEBSITE_ROOT` to the `integration.publishing.service.gov.uk`.

The code will update the destination, while ensuring the Host header is set as appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,16 @@ You can use the following environment variables to configure the tests:
   * Default: Blank
   * Password of a user with a Signon account in the environment the tests are being run in.
 
+### Spoof target domain
+
+Set the `SPOOF_TARGET_DOMAIN` environment variable to run tests against a alternative URL, while
+maintaining the `Host` header as set by the `GOVUK_WEBSITE_ROOT` environment variable.
+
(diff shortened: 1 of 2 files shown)


surminus:
This may fail because of SNI, so we need to investigate if the Ruby library supports doing what we wish to do.

boffbowsh:
It took me a while to work out what this was doing, I think maybe `CUSTOM_DESTINATION_URI` isn't the best name. Maybe `SPOOF_HOST_HEADER` or similar?

surminus:
I updated to `SPOOF_TARGET_DOMAIN` as what this env var is setting is the target URL, whereas the code sorts out the host header.

PR #309, opened 2017-12-11
Add support for bouncer APP_DOMAIN_INTERNAL

afda16:
On AWS some of the end points are internal facing only and are
configured with an internal domain, for instance, Bouncer. Smokey
tests for these endpoints are failing because the monitoring box
is trying to use the public domain specified in GOVUK_APP_DOMAIN.

In this change we add a new environment variable, GOVUK_APP_DOMAIN_INTERNAL,
that, if configured, will be used to create some of the service URLs.

In this case we applied the internal domain to Bouncer.

### DIFF
--- a/features/support/base_urls.rb
+++ b/features/support/base_urls.rb
@@ -2,6 +2,10 @@ def app_domain
   ENV["GOVUK_APP_DOMAIN"] || "integration.publishing.service.gov.uk"
 end
 
+def app_domain_internal
+  ENV["GOVUK_APP_DOMAIN_INTERNAL"]
+end
+
 def signon_base_url
   application_base_url('signon')
 end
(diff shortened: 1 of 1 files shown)


deanwilson:
Merge away

features/support/base_urls.rb
```diff
@@ -2,6 +2,10 @@ def app_domain
   ENV["GOVUK_APP_DOMAIN"] || "integration.publishing.service.gov.uk"
 end
 
+def app_domain_internal
+  ENV["GOVUK_APP_DOMAIN_INTERNAL"] || ""
```
surminus: I'm not sure if the `|| ""` is required. 

If it's not defined then it will return `nil`, and you could put that (`.nil?`) in the tertiary operator below.

afda16: Updated

PR #310, opened 2017-12-11
Allow setting environment from custom application

afda16:
On AWS we want to run tests for the publishing and stack specific
domains. We also need to add new variables to have the tests working.

The smokey tests are called from Jenkins, as a job, via `jenkins.sh`
script, and are also running as a service in the monitoring server. In
both cases environment variables are first loaded from the environment
or from a `/etc/smokey.sh` file, and then set with `govuk_setenv default`.

This makes dificult to set up the right environment in first place and
have tests running with more than one setting, so this change allows us
to change the default `govuk_setenv` to load the environment from several
Smokey applications.

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -10,6 +10,10 @@ if [ -z $MYTASK ]; then
   MYTASK="test:production"
 fi
 
+if [ -z $GOVUK_SETENV_APP ]; then
+  GOVUK_SETENV_APP=default
+fi
+
 # This removes rbenv shims from the PATH where there is no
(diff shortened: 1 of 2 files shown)


issyl0:
What would the expected values of `GOVUK_SETENV_APP` be apart from `default`?

afda16:
Something like 'smokey-publishing' to use setting for the publishing domain, and 'smokey-stack' to use settings for the stack domain

surminus:
I do not believe this required for the time-being, so I am going to close it off.

PR #311, opened 2017-12-13
Disable Smokey sitemap testing on integration

tijmenb:
This test is failing a lot on integration.

Work is planned to fix it in https://trello.com/c/DmzPTudf


MatMoore:
I'll add a note to reenable this one we've fixed the issue in the trello card.

PR #312, opened 2017-12-13
Check for unique link in search results feature

brenetic:
We want to make sure that the returned search results are unique and have previously done this by using the title of a document. In practice our content does not always have unique titles so swap this to use links instead.

### DIFF
--- a/features/search.feature
+++ b/features/search.feature
@@ -17,6 +17,7 @@ Feature: Search
     And the search results should be unique
 
   @high
+  @notintegration
   Scenario: check search results for universal credit
     Given I am testing through the full stack
     And I force a varnish cache miss


features/step_definitions/search_steps.rb
```diff
@@ -15,7 +15,7 @@
 And /^the search results should be unique$/ do
   results = []
   page.all(".results-list li a").each_with_index do |item, idx|
-    results << item.text + page.all(".results-list li p")[idx].text
+    results << item[:href] + page.all(".results-list li p")[idx].text
```
suzannehamilton: If we're comparing the URLs, do we still need to compare the descriptions? (I'm guessing that's what `page.all(".results-list li p")[idx].text` does.)

MatMoore:
Actually I've just noticed that the original implementation is supposed to be checking the combination of title + description. Do we know why this didn't work? I had assumed the problem was two pages with the same title but they do have different descriptions.

brenetic:
`Universal Credit statisticsOfficial experimental statistics on the number of people and households on Universal Credit, starts and claims.` appeared twice :(

brenetic:
@MatMoore So it wasn't actually failing on `Universal Credit` but rather `Universal Credit statistics`

brenetic:
@MatMoore As discussed, we are going to skip running this test on Integration for now.

chrisroos:
Thanks for getting this merged, folks. Fingers crossed Smokey is a bit greener in integration from now on :-)

PR #316, opened 2017-12-19
Add missing "?" in ternary operator

surminus:


issyl0:
D'oh. I should have picked that up. I was too focused on the 🐙 emoji choice.

PR #324, opened 2018-01-17
Change the handling of base URLs

cbaines:
Previously a rather awkward list of URLs was used. This replaces that
list with using Plek, and a hash of default paths. I don't think all
default paths are used, but just include all of the ones specified
previously for now.

This uses the new external_url_for method in Plek 2.1.0 to find
external URLs for services. Some changes to the step definitions are
made here to explicitly use external URLs in some circumstances.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem "plek", "1.11.0"
+gem "plek"
 gem "rake", "0.9.2.2"
 gem "rspec", "2.11.0"
 gem "minitest", "5.8.4"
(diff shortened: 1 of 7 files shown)


cbaines:
These changes are hopefully an improvement, although I haven't tested them locally, as I don't think it's possible to access the internal applications on Integration anymore (as the Smokey README suggests).

This also depends on the yet unreleased 2.1.0 release of Plek https://github.com/alphagov/plek/pull/56

cbaines:
> Looks good, and I think it'll fix some smokey tests for bouncer that have been failing for a while.

Great, thanks for reviewing Murray :)

PR #325, opened 2018-01-19
Remove redundant/failing tests for asset content-types

chrisroos:
These tests were added in 3725f865d22340698fad39ae7f3cca20d0d2ecfa as part of our investigation into the missing content-type issue captured in https://github.com/alphagov/asset-manager/issues/238. The cause of this problem was fixed in https://github.com/alphagov/asset-manager/pull/343 which rendered these tests redundant.

Not only are these tests redundant but they started failing when we merged https://github.com/alphagov/asset-manager/pull/418 to ensure we're setting the correct content-type for a number of additional file types.

### DIFF
--- a/features/assets.feature
+++ b/features/assets.feature
@@ -5,15 +5,3 @@ Feature: Assets
     Given I am testing "assets"
     When I request "/__canary__"
     Then I should get a 200 status code
-
-  @normal
-  Scenario: Assets with an explicit Content-Type are served with the correct Content-Type
-    Given I am testing "assets"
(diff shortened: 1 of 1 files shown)


boffbowsh:
Can we update them instead to cover the correct content type please? I think this test is still worthwhile to capture regressions.

chrisroos:
> Can we update them instead to cover the correct content type please? I think this test is still worthwhile to capture regressions.

@boffbowsh. Sure thing. I'll open a new PR although I'm not entirely convinced how useful they given all the tests I added in https://github.com/alphagov/asset-manager/pull/418.

chrisroos:
@boffbowsh - I've opened https://github.com/alphagov/smokey/pull/326 to replace this. Can you take a look, please?

PR #326, opened 2018-01-19
Fix failing tests for asset content-types

chrisroos:
These tests started failing when we merged https://github.com/alphagov/asset-manager/pull/418 to set the correct content-type for a number of additional file types.

These tests were added in 3725f865d22340698fad39ae7f3cca20d0d2ecfa as part of our investigation into the missing content-type issue captured in https://github.com/alphagov/asset-manager/issues/238. The cause of this problem was fixed in https://github.com/alphagov/asset-manager/pull/343 but we're keeping these tests around to help catch regression in future.

### DIFF
--- a/features/assets.feature
+++ b/features/assets.feature
@@ -7,13 +7,13 @@ Feature: Assets
     Then I should get a 200 status code
 
   @normal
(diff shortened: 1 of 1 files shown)


chrisroos:
Thanks for the review, @boffbowsh!

PR #328, opened 2018-01-26
Remove testing for policy firehose

tijmenb:
This removes a test for a feature removed in https://github.com/alphagov/policy-publisher/pull/221. Also removes a duplicate of the "check policy page loads" test at L16.

We're no longer testing that the dropdown for "empty values", but as far as I know there aren't any finders with document type, person or policy dropdowns.

https://trello.com/c/YDda1PqI

### DIFF
--- a/features/finder_frontend.feature
+++ b/features/finder_frontend.feature
@@ -4,7 +4,6 @@ Feature: Finder Frontend
   Background:
     Given I am testing through the full stack
     And I force a varnish cache miss
-    And I am in the "A" group for "PolicyFinderTest" AB testing
 
   @normal
   Scenario: check people page loads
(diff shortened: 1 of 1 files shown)


features/finder_frontend.feature
```diff
@@ -19,16 +19,6 @@ Feature: Finder Frontend
...
 
-  @normal
-  Scenario: check policy alpha loads
-    When I visit "/government/policies/all"
-    Then I should see "All policy content"
-    And I should see an input field to search
-    And I should see an open facet titled "Document type" with non-blank values
-    And I should see a closed facet titled "Organisation" with non-blank values
-    And I should see a closed facet titled "Policy" with non-blank values
-    And I should see a closed facet titled "People" with non-blank values
```
leenagupte: I think these three lines should be added to another test. They were added in commit b40f5d01b9ea900e255a2a56aa7cde3638c930f9 to highlight when things (people etc) are missing from rummager.

cc @MatMoore

PR #329, opened 2018-02-07
Fix the ambiguity of the I am testing step

cbaines:
Add a $ at the end to separate this from the I am testing "..."
internally step.


cbaines:
This is causing a problem in integration.

```
Ambiguous match of "I am testing "asset-manager" internally":

features/step_definitions/smokey_steps.rb:3:in `/^I am testing "(.*)"/'
features/step_definitions/smokey_steps.rb:11:in `/^I am testing "(.*)" internally/'
```
https://deploy.integration.publishing.service.gov.uk/job/Smokey/243/console

PR #330, opened 2018-02-07
Don't directly access Whitehall Frontend

cbaines:
This test is confusing, as it seems to suggest it's testing through
the full stack, including varnish, but then attempts to access
whitehall-frontend directly.

This doesn't work on the new AWS Integration, so lets change this to
just visit the URL via the normal route.

### DIFF
--- a/features/whitehall.feature
+++ b/features/whitehall.feature
@@ -100,7 +100,7 @@ Feature: Whitehall
     Given I am benchmarking
     And I am testing through the full stack
     And I force a varnish cache miss
-    When I visit "/government/how-government-works" on the "whitehall-frontend" application
+    When I visit "/government/how-government-works"
     Then the elapsed time should be less than 2 seconds
 
   @normal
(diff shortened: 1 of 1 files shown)


features/whitehall.feature
```diff
@@ -100,7 +100,7 @@ Feature: Whitehall
     Given I am benchmarking
     And I am testing through the full stack
     And I force a varnish cache miss
-    When I visit "/government/how-government-works" on the "whitehall-frontend" application
+    When I visit "/government/how-government-works"
```
h-lame: Should we do the same to the "Whitehall offers a world location API" test further down?  Doesn't seem to make much sense that we check that against admin boxes when it's a public API (isn't it?)

cbaines: Yeah, I ignored that as I think it's working (and should generally work), but yes, that test is similarly confusing, so I'm happy to change it as well.

cbaines: Done in 58e3ed1

h-lame: :+1: - I guess the real test is shipping this and seeing that it still passes.

PR #336, opened 2018-02-08
Add smokey tests for the different email sign up journeys

1pretz1:
There is one end to end test which is from "/foreign-travel-advice/turkey"
to the sign up confirmation page. The rest are from different entry points and stop
after getting to the email frequency page, as from then, the journeys follow a standardised route.
Also removed 'I click ...' step as none of the features were using it.

### DIFF
--- a/features/step_definitions/smokey_steps.rb
+++ b/features/step_definitions/smokey_steps.rb
@@ -165,12 +165,6 @@ Then /^I should be at a location path of "(.*)"$/ do |location_path|
   end
 end
 
-When /^I click "(.*?)"$/ do |link_text|
-  link_href = Nokogiri::HTML.parse(@response.body).at_xpath("//a[text()='#{link_text}']/@href")
-  link_href.should_not == nil
-  step "I visit \"#{link_href.value}\""
-end
-
 When /^I try to post to "(.*)" with "(.*)"$/ do |path, payload|
   @response = post_request "#{@host}#{path}", :payload => "#{payload}"
 end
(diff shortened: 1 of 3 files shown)


1pretz1:
Hey, I've discussed this with Kevin and Ruben and seen as though the path standardises after the frequency page. We thought there wouldn't be much point in going end to end in every scenario as if the test passes it would pass for every scenario and vise-versa for if it failed  ?

features/step_definitions/smokey_steps.rb
```diff
@@ -158,9 +158,13 @@ def should_see(text)
...
 When /^I click "(.*?)"$/ do |link_text|
-  link_href = Nokogiri::HTML.parse(@response.body).at_xpath("//a[text()='#{link_text}']/@href")
+  if @response
+    link_href = Nokogiri::HTML.parse(@response.body).at_xpath("//a[text()='#{link_text}']/@href")
+  elsif body
+    link_href = Nokogiri::HTML.parse(page.body).at_xpath("//a[text()='#{link_text}']/@href")
+  end
   link_href.should_not == nil
-  step "I visit \"#{link_href.value}\""
+  step  "I visit \"#{link_href.value}\""
```
rubenarakelyan: Minor nitpick to remove this extra space

gpeng:
You could do with tidying up the commits here before we merge it. Give me a shout and I can give you a hand 👋 ◀️

1pretz1:
Okay, thankyou! After i deploy the other change 'ignore notify email' ill give you a holla.

PR #341, opened 2018-03-08
Add scenario to ensure draft-assets is protected by signon

chrisroos:
I noticed that visiting draft-assets is currently redirecting us to the internal signon URL instead of the external URL.

Adding this scenario should help us avoid similar problems in future.


chrisroos:
Note that this shouldn't be merged until the fix in https://github.com/alphagov/asset-manager/pull/512 has been merged/deployed.

chrisroos:
Thanks, @floehopper.

As mentioned, I've had to update this test after noticing a problem when trying it against integration now that https://github.com/alphagov/asset-manager/pull/512 has been merged. Can you take another look at this PR, please?

PR #343, opened 2018-03-13
Change the asset whitehall uses for a smokey test

kevindew:
The previous asset was returning a 302 which seems to be related to the
parent document it is associated with being unpublished with redirect.
I'm not 100% sure if that behavious is correct or not so raised it with
the asset-management team.

I have though switched this to using a asset of a document that is
published which cuts out a level of ambiguity.

### DIFF
--- a/features/whitehall.feature
+++ b/features/whitehall.feature
@@ -115,7 +115,7 @@ Feature: Whitehall
   @normal
   Scenario: Whitehall assets are served
     Given I am testing through the full stack
-    When I request "/government/uploads/system/uploads/attachment_data/file/32409/11-944-higher-education-students-at-heart-of-system.pdf"
+    When I request "/government/uploads/system/uploads/attachment_data/file/618167/government_dietary_recommendations.pdf"
     Then I should get a 200 status code
 
   @normal


chrisroos:
I think the failure has highlighted another problem with this test. It was previously succeeding because it's been redirecting to a page on gov.uk that returns a 200 response, and not because the attachment was actually being served. I think the test needs to be changed so that it asserts we receive a 200 response directly, and not via a redirect.

chrisroos:
> I think the failure has highlighted another problem with this test. It was previously succeeding because it's been redirecting to a page on gov.uk that returns a 200 response, and not because the attachment was actually being served. I think the test needs to be changed so that it asserts we receive a 200 response directly, and not via a redirect.

I'm investigating how easy it is to do this.

kevindew:
@chrisroos That's interesting. I can't see any signs of anything recent changing on it as well for the 200 - from whitehall admin it looks like https://www.gov.uk/government/publications/higher-education-students-at-the-heart-of-the-system should be redirecting but isn't 🤔 

Anyway it doesn't look too hard to turn off redirects since this used the RestClient library.

chrisroos:
I've opened PR #345 in an attempt to avoid false positives with this test in future.

chrisroos:
> from whitehall admin it looks like https://www.gov.uk/government/publications/higher-education-students-at-the-heart-of-the-system should be redirecting but isn't 🤔

Yeah, I noticed that too. I'm relatively confident that's a separate issue so haven't investigated.

PR #344, opened 2018-03-13
Update the gemfile and dependencies

kevindew:
This makes the gemfile consistent in it's dependency definitions and does a
bundle update.

Example of a passing build: https://deploy.staging.publishing.service.gov.uk/job/Smokey/8592/console


cbaines:
> This makes the gemfile consistent in it's dependency definitions and does a
> bundle update.

This seems to add a bit more information on top of the commit message, it might be good to add it?

kevindew:
@cbaines thanks done! not sure why I didn't have that there in the first place.

PR #345, opened 2018-03-13
Ensure the asset url requested is not redirected

chrisroos:
A recent failure in this test highlighted that it hadn't been passing for the right reason: the attachment being requested belonged to an unpublished document and, because [RestClient follows redirects by default][1], we were actually asserting that the resulting document page was a 200.

This test was originally added in c9d72358e97fbdd92c7fe99517fac4b874870dee but there's no indication in the commit note or associated Pivotal Tracker story that we were expecting anything other than to receive the actual attachment directly.

By not following redirects when we make the request for the asset we should avoid similar false positives in future.

[1]: https://github.com/rest-client/rest-client#redirection

### DIFF
--- a/features/step_definitions/smokey_steps.rb
+++ b/features/step_definitions/smokey_steps.rb
@@ -38,14 +38,16 @@ When /^I go to the "([^"]*)" landing page$/ do |app_name|
   visit_path application_external_url(app_name)
 end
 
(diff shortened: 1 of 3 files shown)


chrisroos:
Thanks for reviewing, @kevindew. I'll get this merged.

PR #351, opened 2018-03-27
Fix failing Whitehall attachments scenario

chrisroos:
Whitehall [PR 3918][1] added functionality to redirect requests for attachments from gov.uk to the asset host. This functionality caused the "Whitehall assets are served" scenario to fail because it was expecting to be able to request an asset without it being redirected.

The fix is to assert that Whitehall attachments _are_ redirected from gov.uk to the asset host, and to separately assert that Whitehall attachments are successfully served from the asset host.

[1]: https://github.com/alphagov/whitehall/pull/3918

### DIFF
--- a/features/whitehall.feature
+++ b/features/whitehall.feature
@@ -113,8 +113,14 @@ Feature: Whitehall
     Then the elapsed time should be less than 2 seconds
 
   @normal
-  Scenario: Whitehall assets are served
+  Scenario: Whitehall assets are redirected to the asset host
     Given I am testing through the full stack
+    When I request "/government/uploads/system/uploads/attachment_data/file/618167/government_dietary_recommendations.pdf"
+    Then I should be redirected to the asset host
+
+  @normal
+  Scenario: Whitehall assets are served from the asset host
+    Given I am testing "assets-origin"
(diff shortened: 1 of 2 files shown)


features/whitehall.feature
```diff
@@ -113,8 +113,14 @@ Feature: Whitehall
     Then the elapsed time should be less than 2 seconds
 
   @normal
-  Scenario: Whitehall assets are served
+  Scenario: Whitehall assets are redirected to the asset host
     Given I am testing through the full stack
+    When I request "/government/uploads/system/uploads/attachment_data/file/618167/government_dietary_recommendations.pdf"
+    Then I should be redirected to the asset host
```
thomasleese: Do we need to check the filename for the redirect URL matches the given one?

thomasleese: Oh I guess that's covered by checking for the 200 response later on.

chrisroos:
Thanks, @thomasleese!

features/whitehall.feature
chrisroos: Yeah, I think it's good enough to check the host given that the next scenario checks that we can successfully request the attachment. I considered removing the redirect test given that it's well tested within Whitehall but decided that it shouldn't do any harm to test it here too.

PR #353, opened 2018-04-10
Ensure the related content appears on Travel Advice

boffbowsh:
With the current Taxonomy changes, it would be easy to break
the hardcoded related content for Travel Advice. This checks
for the existance of the link to the "About" page.


boffbowsh:
`git add .` 🤦‍♂️

PR #355, opened 2018-04-12
[DO NOT MERGE] Check which frontend app renders the response

barrucadu:
A demo of checking which frontend app renders the response.  This is useful to ensure that we don't accidentally lose coverage when pages migrate between apps.  Approach copied from https://github.com/alphagov/publishing-e2e-tests/blob/e79bf84e2c3e5f27e1396ca498c0a7bd278f72fa/spec/support/frontend_helpers.rb#L2-L7


barrucadu:
I'll close this, as it was just to check that the approach worked and share it with others.  As the Trello card links to this, it won't be lost.

PR #357, opened 2018-04-19
Fix data.gov.uk smokey test for .rdf links

rossjones:
This URL now returns a 301 rather than a 200


rossjones:
Closed as the checks are following redirects and they do eventually resolve to a 200 - so this PR is incorrect.

PR #360, opened 2018-04-23
Add a test for the benefits info page

barrucadu:
Request a page served by info-frontend and look for some strings.

### DIFF
--- a/features/info_frontend.feature
+++ b/features/info_frontend.feature
@@ -1,0 +1,10 @@
+Feature: Info frontend
+  Background:
+    Given I am testing through the full stack
+    And I force a varnish cache miss
+
+  @normal
+  Scenario:
+    When I visit "/info/browse/benefits"
+    Then I should see "Benefits"
+     And I should see "Unique pageviews"


features/info_frontend.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Manuals frontend
```
kevindew: Should this be "Info frontend"?

features/info_frontend.feature
```diff
@@ -0,0 +1,13 @@
...
+    And I force a varnish cache miss
+
+  @normal
+  Scenario:
+    When I visit "/info/browse/benefits"
+    Then I should see "Benefits"
+     And I should see "Unique pageviews"
+     And I should see "Searches started from this page"
+     And I should see "Problem reports"
+     And I should see "Users left by searching for"
```
kevindew: Should there be a step here where we check the meta tag that it is rendered by "info-frontend"?

I'm also not sure we need to check each of these areas exist. I'd have thought that so long as we have "Benefits" that would be sufficient - rather than needing to test that features provided by info-frontend are present.

features/info_frontend.feature
barrucadu: Yes it should

features/info_frontend.feature
barrucadu: Could do, but none of the other tests do that at the moment, so it would be a bit inconsistent.

kevindew: Sure, If I remember right that was something we identified as an improvement in that we're testing info-frontend but we don't actually know if info-frontend is serving. 

Could be fine as a separate story though to add this to all ones that are intended to test a particular app.

barrucadu: I'll make a ticket for that.

PR #361, opened 2018-04-23
Check that smart answers strip postcodes

barrucadu:
There are no tests for static, so let's look for a meta tag set in https://github.com/alphagov/static/blob/master/app/views/govuk_component/analytics_meta_tags.raw.html.erb.

### DIFF
--- a/features/smartanswers.feature
+++ b/features/smartanswers.feature
@@ -132,3 +132,8 @@ Feature: Smart Answers
       | /benefit-cap-calculator/y/yes/no/none/bereavement/1000.0/1000.0/single/EH99%201SP | Outside London benefit cap |
       | /landlord-immigration-check/y/SW1A%202AA                                          | Is the person renting      |
       | /landlord-immigration-check/y/EH99%201SP                                          | check in England           |
+
+  @normal
+  Scenario: Excluding personal information
+    When I visit "/marriage-abroad/y"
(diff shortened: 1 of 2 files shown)


tijmenb:
Incidentally, I'm moving the analytics meta tag thing from static to govuk_publishing_components (https://github.com/alphagov/govuk_publishing_components/pull/278), so this won't test static after that.

Is there a particular thing you're after by testing static?

barrucadu:
Not really, it's just that there's no Smokey coverage of static things.  But perhaps it's not necessary.

tijmenb:
Ah right. Static also provides the layout of the page, if it isn't available the page wouldn't render at all.

barrucadu:
Ok, so by having smokey tests which (eg) check for a string in a page, enough is being done to test static?

tijmenb:
I do like it if we'd test that PII feature though, especially since we're moving it from static -> govuk_publishing_components!

tijmenb:
Maybe move it to the smart answers test though.

barrucadu:
I'll rename the file to make it a bit clearer then.

tijmenb:
👍

tijmenb:
Well this turned out to be a very sensible test to add, because I managed to remove the PII stripping in https://github.com/alphagov/smart-answers/pull/3506 🤕. 

Because smokey failed @vanitabarrett caught this on integration!

PR #362, opened 2018-04-24
Check that the contact form renders

barrucadu:
Hit /contact/govuk and check the page renders.

---


features/feedback.feature
```diff
@@ -0,0 +1,9 @@
+@disabled_in_icinga
```
thomasleese: I'm probably just missing some context since I've been away, but why is this?

barrucadu: Oh yeah, that shouldn't have stayed there.

PR #365, opened 2018-05-31
Add coverage for search dependent pages and feeds

steventux:
https://trello.com/c/LI3rZlO6/196-look-at-the-list-of-pages-powered-by-search-and-add-smokey-tests

Adds tests to verify that various pages depending on Rummager searches are functioning and return results.


features/search_dependents.feature
```diff
@@ -0,0 +1,27 @@
+Feature: Pages depending on search
+
+  @normal
+  Scenario Outline: check results for pages relying on search
+    Given I am testing through the full stack
+    And I force a varnish cache miss
+    When I visit "<Path>"
+    Then I should see search results at css "<Search results selector>"
+
+    Examples:
```
tijmenb: I'm wondering if these selectors are too brittle (eg it's very likely that `app-c-topic-list` will change soon). Aren't most of these things tested in their own smokey tests?

PR #368, opened 2018-06-18
Test for input sanitisation

jonodrew:
This PR is for three smokey tests. It tests a basic XSS attack against the following:

- search
- feedback
- finder

### DIFF
--- a/features/finder_frontend.feature
+++ b/features/finder_frontend.feature
@@ -57,3 +57,8 @@ Feature: Finder Frontend
   Scenario: check that advanced search returns results
     When I visit "/search/advanced?topic=/education&group=news_and_communications"
     Then I should see filtered documents
+
+  @high
+  Scenario: check malicious code does not execute
+    When I visit "/government/organisations/hm-revenue-customs/contact?keywords=<script>alert(document.cookie);</script>"
+    Then I should see "<script>alert(document.cookie)" in the results section
--- a/features/feedback.feature
+++ b/features/feedback.feature
@@ -6,3 +6,8 @@ Feature: Feedback
   Scenario: The "Contact GOV.UK" page
     When I visit "/contact/govuk"
     Then I should see "Contact GOV.UK"
+
+  Scenario: malicious actor inputs code to carry out XSS attack
+    When I visit "contact/govuk"
+    And I input malicious code in the email field
+    Then I see the code returned in the page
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 webrat.log
-log/*.log
\ No newline at end of file
+log/*.log
+testyCucumber.sh
```
elliotcm: Probably don't need this anymore?

testyCucumber.sh
```diff
@@ -0,0 +1,10 @@
+GOVUK_DRAFT_WEBSITE_ROOT=https://draft-origin.integration.publishing.service.gov.uk \
+SIGNON_EMAIL="<email-address>" \
+SIGNON_PASSWORD="<password>" \
+AUTH_USERNAME="<username>" \
+AUTH_PASSWORD="<password>" \
+bundle exec cucumber features/email_sign_up.feature \
+--format pretty \
+--tags ~@pending \
+--tags ~@local-network \
+--tags ~@notintegration
```
elliotcm: This file looks like it should be removed.

features/finder_frontend.feature
```diff
@@ -57,3 +57,8 @@ Feature: Finder Frontend
   Scenario: check that advanced search returns results
     When I visit "/search/advanced?topic=/education&group=news_and_communications"
     Then I should see filtered documents
+
+  @high
+  Scenario: check malicious code does not execute
+    When I visit "/government/organisations/hm-revenue-customs/contact?keywords=<script>alert(document.cookie);</script>"
+    Then I should see "<script>alert(document.cookie)" in the result-info class
```
elliotcm: Do we need to say "in the result-info class" here?  Would be nicer to decouple the step text from the specific implementation of the page, e.g. saying "results section" (if that's what it is)

features/form_input.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Form input
+
+@high
+Scenario: Malicious actor inputs executable code to carry out XSS attack
+  Given I am a malicious actor
+  When I visit "/foreign-travel-advice/turkey"
+  And I click on the link "Get email alerts"
+  Then I should see "Create subscription"
+  When I click on the button "Create subscription"
```
elliotcm: Quite a few steps here, probably best to move some of the irrelevant details into a couple of simpler steps e.g. "When I add a malicious script to the subscription signup form"

elliotcm:
Would also be nice to squash and/or rename some of the commits, especially the WIP one.

elliotcm:
(see here for PR styleguide https://github.com/alphagov/styleguides/blob/master/pull-requests.md)

PR #371, opened 2018-07-02
Run bundle update

cbaines:


cbaines:
Tested in Staging, seems to work https://deploy.staging.publishing.service.gov.uk/job/Smokey/9734/console

PR #372, opened 2018-07-02
Update gem major versions

cbaines:
Also only specify the major version in the Gemfile, as the exact
version is specified in the Gemfile.lock.


Gemfile
```diff
@@ -1,12 +1,12 @@
...
-gem "rake", "~> 0.9"
-gem "rest-client", "~> 1.8"
-gem "rspec", "~> 2.11"
+gem "capybara", "~> 3"
+gem "cucumber", "~> 3"
+gem "minitest", "~> 5"
+gem "nokogiri", "~> 1"
+gem "plek", "~> 2"
+gem "poltergeist", "~> 1"
+gem "rake", "~> 11"
```
leenagupte: 😱

PR #399, opened 2018-07-24
Specify port for proxy

rubenarakelyan:
This commit specifies a set port for the proxy to override the default of `8080` (which is already in use on the machines on which Smokey runs).

Trello: https://trello.com/c/80OiCEtH/308-more-reliable-smokey-tests-on-aws-environments


features/support/env.rb
```diff
@@ -32,7 +32,7 @@
 
 # Set up proxy server (used to manipulate HTTP headers etc since Selenium doesn't
 # support this)
-server = BrowserMob::Proxy::Server.new("./bin/browsermob-proxy")
+server = BrowserMob::Proxy::Server.new("./bin/browsermob-proxy", port: 3222)
```
thomasleese: Can we make it randomly pick a port between a range? Might help if someone tries to run Smokey twice?

PR #419, opened 2018-08-23
Disable failing scenario

steventux:
This scenario has been failing for a week or so with the error:
`Failed to open TCP connection to service.gov.uk:443 (Connection refused...`


timblair:
This is due to a change in the [Anycast IP addresses](https://github.com/alphagov/govuk-dns-config/pull/129). We've just discovered that the new IPs Fastly gave us are non-TLS ones, hence we started getting connection errors.

benthorner:
Ah, OK. Thanks for helping with my understanding here - I wasn't sure who to ask :-)

PR #455, opened 2018-11-08
Add tests for content data admin

cbaines:
Trello: https://trello.com/c/IDm7nzwL/728-5-add-end-to-end-integration-tests

### DIFF
--- a/features/content_data_admin.feature
+++ b/features/content_data_admin.feature
@@ -1,0 +1,19 @@
+Feature: Content Data Admin
+  Tests for the Content Data Admin application, which provides
+  publishers with data about the content they manage
+
+  @normal
+  Scenario: Can access the Content Data Admin index page
+    When I go to the "content-data-admin" landing page
+    And I try to login as a user
+    And I go to the "content-data-admin" landing page
+    Then I should see "Content data"
+    And I should see "Log out"
+
+  @normal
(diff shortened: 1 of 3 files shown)


features/content_data_admin.feature
```diff
@@ -0,0 +1,22 @@
+Feature: Content Data Admin
+  Tests for the Content Data Admin application, which provides
+  publishers with data about the content they manage
+
+  # TODO: Doesn't work yet, as the Smokey user doesn't have an
```
pmanrubia: @cbaines what do we have to do to have an organisation for Smokey?

cbaines: > @cbaines what do we have to do to have an organisation for Smokey?

We could set an organisation, I think it's as simple as just making the change in the Signon user interface. The tricky bit is making sure that doesn't have any unexpected effects on the other tests.

cbaines: > We could set an organisation, I think it's as simple as just making the change in the Signon user interface. The tricky bit is making sure that doesn't have any unexpected effects on the other tests.

I've made this change now, and I've updated the comment. Even though it should work now, most of the time the test will still fail due to the slowness of the index page.

cbaines: With the index page speed improvements, I've marked both tests as `@normal` now.

PR #466, opened 2018-12-19
Change better-business-finance example

deborahchua:
The example currently being used has been unpublished in `specialist-finder`.
The document replacing it looks like it's unlikely to change in the near future as it links to a directory of approved finance suppliers.


sihugh:
🚀

PR #471, opened 2019-02-07
introduce new AWS staging enviroment

fredericfran-gds:


rubenarakelyan:
We now have this for AWS staging and production via https://github.com/alphagov/smokey/pull/479

PR #493, opened 2019-04-02
Add a Smokey test for data.gov.uk dataset count

thomasleese:
This checks that the dataset count in CKAN closely matches the dataset count in the data.gov.uk Find frontend. This is because sometimes the syncing process can fail between CKAN and Find, and this test should alert us to that fact quickly.

[Trello Card](https://trello.com/c/TGd6sSbe/904-add-a-smokey-check-for-the-number-of-datasets-in-datagovuk-find)

### DIFF
--- a/features/data_gov_uk.feature
+++ b/features/data_gov_uk.feature
@@ -28,3 +28,21 @@ Feature: Data.gov.uk
     And I force a varnish cache miss
     When I request "/"
     Then I should see "Data publisher"
+
+  @high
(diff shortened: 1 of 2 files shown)


thomasleese:
@elliotcm Interesting. I had thought about an absolute check, but I thought it might be annoying if we keep having to update the number. But let's put one in and review if it turns out to be annoying.

thomasleese:
@elliotcm Thanks for that, I've pushed up a commit to add an absolute check as well.

PR #524, opened 2019-05-13
Improvements to documentation

ChrisBAshton:
## Java instructions

I was surprised to run `bundle install` and see a system popup about a missing Java SDK, so it's a good idea to be explicit that Java is required.

I also ended up registering for an Oracle account, not knowing about the OpenJDK alternative, so have documented an advised installation process.

## HTTP `RestClient::RequestFailed` error

Whilst [it is documented in the Developer Docs](https://docs.publishing.service.gov.uk/manual/alerts/high-priority-tests.html), this is a common enough error that I believe it's worth a place in the README. The error happens whenever you cancel a running test suite, which you're likely to do if you notice the first test has failed.

## Environments and triggers
Current README doesn't make it clear a) which environments the tests run against, and b) what triggers the tests. I've attempted to clarify it here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ After that, it's a standard Ruby setup:
 bundle install
 ```
 
-### Running the tests
+## Running the tests
 
 Run the suite with:
 
@@ -87,6 +87,24 @@ You can use the following environment variables to configure the tests:
 * `RATE_LIMIT_TOKEN`
   * Default: Blank
   * A token used to bypass the default rate limiting.
+
+### HTTP status code failure
+
+A common test failure is `HTTP status code 550 (RestClient::RequestFailed)`. This is a result of the BrowserMob Proxy java process running as part of a previously aborted smokey-loop and the new smoke tests cannot start a new proxy.
+
+It's necessary to kill the existing java process (replace process numbers as appropriate).
+
+```sh
+$ ps -ef | grep java
+> smokey    6385  6380 26 14:58 ?        00:00:54 java -Dapp.name=browsermob-proxy -Dbasedir=/opt/smokey -jar /opt/smokey/lib/browsermob-dist-2.1.4.jar --port 3222
+$ sudo kill -9 6385
+```
+
+You can even set up an alias in your `~/.bash_profile`:
+
+```sh
+alias killbrowsermob="ps xu | grep browsermob-proxy | grep -v grep | awk '{ print $2 }' | xargs kill -9"
+```
 
 ### Spoofing the target domain
 
--- a/README.md
+++ b/README.md
@@ -13,6 +13,20 @@ files to describe single applications (eg
 [`whitehall`](https://github.com/alphagov/whitehall),
 [`frontend`](https://github.com/alphagov/frontend)) or [cross-application behaviour](features/gov_uk.feature).
 
+### Installation
+
+Smokey requires Java to be installed, because of its [use of the BrowserMob Proxy](#use-of-browsermob-proxy). Note that if you're using the Development VM then Java is already installed.
+
+If you're not using the VM, or want to run Smokey on your host Mac, run `brew cask install java`.
+
+If you don't have Homebrew installed, or are not using a Mac, you can [download the Java JDK from the OpenJDK website](https://openjdk.java.net/).
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,6 +13,16 @@ files to describe single applications (eg
 [`whitehall`](https://github.com/alphagov/whitehall),
 [`frontend`](https://github.com/alphagov/frontend)) or [cross-application behaviour](features/gov_uk.feature).
 
+### Installation
+
+Smokey requires Java to be installed, because of its [use of the BrowserMob Proxy](#use-of-browsermob-proxy). [Download Java from the Oracle website](https://www.oracle.com/technetwork/java/javase/downloads/index.html) (you'll need to create an Oracle account).
```
cbaines: It's probably worth noting that these instructions are for MacOS right? And wouldn't apply if you're doing this within the Ubuntu development-vm?

README.md
```diff
@@ -6,14 +6,26 @@ applications within the GOV.UK stack.
...
 
 The smoke tests are based on [Cucumber](https://cucumber.io/). We use feature
 files to describe single applications (eg
 [`whitehall`](https://github.com/alphagov/whitehall),
 [`frontend`](https://github.com/alphagov/frontend)) or [cross-application behaviour](features/gov_uk.feature).
 
-### Running the tests
+### Installation
+
+Smokey requires Java to be installed, because of its [use of the BrowserMob Proxy](#use-of-browsermob-proxy). [Download Java from the Oracle website](https://www.oracle.com/technetwork/java/javase/downloads/index.html) (you'll need to create an Oracle account). Note that if you're using the Development VM then Java is already installed.
```
timblair: Can we use OpenJDK instead of Oracle's version? That can be installed using `brew cask install java` without the need for an Oracle account.

README.md
```diff
@@ -74,6 +86,18 @@ You can use the following environment variables to configure the tests:
   * Default: Blank
   * A token used to bypass the default rate limiting.
 
+### HTTP status code failure
+
+A common test failure is `HTTP status code 550 (RestClient::RequestFailed)`. This is a result of the BrowserMob Proxy java process running as part of a previously aborted smokey-loop and the new smoke tests cannot start a new proxy.
+
+It’s necessary to kill the existing java process (replace process numbers as appropriate).
```
timblair: Use ASCII single quotes for apostrophes, please :)

README.md
ChrisBAshton: Wish I'd known that! 😄 Will update 👍

README.md
```diff
@@ -6,14 +6,28 @@ applications within the GOV.UK stack.
...
 files to describe single applications (eg
 [`whitehall`](https://github.com/alphagov/whitehall),
 [`frontend`](https://github.com/alphagov/frontend)) or [cross-application behaviour](features/gov_uk.feature).
 
-### Running the tests
+### Installation
+
+Smokey requires Java to be installed, because of its [use of the BrowserMob Proxy](#use-of-browsermob-proxy). Note that if you're using the Development VM then Java is already installed.
+
+If you're not using the VM, [download the Java JDK from the OpenJDK website](https://openjdk.java.net/). Your download will look something like `openjdk-12.0.1_osx-x64_bin.tar.gz`. Untar this file and `sudo mv` it to `/Library/Java/JavaVirtualMachines`. You should now be able to run `java -version`.
```
timblair: Given our [getting started instructions](https://docs.publishing.service.gov.uk/manual/get-started.html#1-install-some-dependencies) include installing rbenv using Homebrew, I'd suggest including the simple installation via that route. Something like this?

```suggestion
If you're not using the VM, or want to run Smokey on your host Mac, then you have two options:

* If you have Homebrew installed, you can install OpenJDK by using `brew cask install java`.
* If not, [download the Java JDK from the OpenJDK website](https://openjdk.java.net/), which will be a file named something like `openjdk-12.0.1_osx-x64_bin.tar.gz`. Untar this file (`tar zxf [FILE]`) and `sudo mv` it to `/Library/Java/JavaVirtualMachines`.

Once you've done this, run `java -version` to check the installation has been successful.
```

ChrisBAshton: Agreed, somehow I glossed over your homebrew suggestion - apologies!

PR #525, opened 2019-05-13
Email default frequency

ChrisBAshton:
Trello card: https://trello.com/c/rUqtUSMG

## What

- Verifies that the business finder email signup is 'No more than once a day' by default.
- Verifies that other existing email signups continue to be 'As soon as they happen' by default.

## Why

In https://trello.com/c/f7DWeg4q, we changed the default value of the business finder email screen to daily. As this touched email-alert-frontend and finder-frontend, we were able to test in isolation but not as a whole, so an integration test is required to check the e2e functionality.

We also want to verify that we haven't broken the default value of existing email signup pages.

### DIFF
--- a/features/step_definitions/email_sign_up_steps.rb
+++ b/features/step_definitions/email_sign_up_steps.rb
@@ -20,3 +20,8 @@ end
 When /^I click on the button "(.*?)"$/ do |button_text|
   click_button button_text
 end
+
+And /^the "(.+)" option should be preselected by default$/ do |frequency|
+  # 'visible: false' because the radio buttons are 'hidden' in order to style them
+  expect(page).to have_css("input[name='frequency'][value='#{frequency}'][checked]", visible: false)
+end
--- a/features/email_sign_up.feature
+++ b/features/email_sign_up.feature
@@ -75,3 +75,12 @@ Feature: Email signup
     And I choose the checkbox "Markets" and click on "Create subscription"
     Then I should see "How often do you want to get updates?"
     And the "immediately" option should be preselected by default
+
+  @normal
+  Scenario: Starting from the business finder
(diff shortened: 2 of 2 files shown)


features/step_definitions/email_sign_up_steps.rb
```diff
@@ -20,3 +20,8 @@
 When /^I click on the button "(.*?)"$/ do |button_text|
   click_button button_text
 end
+
+And /^the "(.+)" option should be preselected by default$/ do |frequency|
+  # 'visible: false' because the radio buttons are 'hidden' in order to style them
+  expect(page).to have_css("input[name='frequency'][value='#{frequency}'][checked]", visible: false)
```
brucebolt: Are the radio buttons definitely 'hidden' in the CSS?  I'm not seeing anything to suggest this on the production site.

ChrisBAshton: Our radio input has this CSS:

```css
.govuk-radios__input {
    cursor: pointer;
    position: absolute;
    z-index: 1;
    top: -2px;
    left: -2px;
    width: 44px;
    height: 44px;
    margin: 0;
    opacity: 0;
}
```

In this case, I think it's the `opacity: 0` that's causing `visible?` to return `false`. The [docs are very vague](https://www.rubydoc.info/github/jnicklas/capybara/Capybara%2FNode%2FElement:visible%3F) ("Not all drivers support CSS, so the result may be inaccurate.") but [other people seem to be able to corroborate this](https://til-engineering.nulogy.com/posts/461fe077bc-capybara-will-skip-invisible-elements-by-default).

brucebolt: I see, didn't spot the `opacity` value is zero.  Seems entirely reasonable to set `visible: false` in that case.

PR #528, opened 2019-05-13
Add s3 CSP test for organograms previews

AlanGabbianelli:
- Check that it is an organogram page by relying on the
  'View the full organogram' text of the button
- Check that we don't get any errors similar to:
  ```
  Refused to connect to
  'https://s3-<organogram bucket>.csv' because it violates the following
  Content Security Policy directive: "connect-src...".
  ```

Co-Authored-By: Ken Tsang <ken.tsang@digital.cabinet-office.gov.uk>

## Reference

https://trello.com/c/wh8pUMOm/982-fix-display-of-organogram-visualisations-on-datagovuk

### DIFF
--- a/features/data_gov_uk.feature
+++ b/features/data_gov_uk.feature
@@ -44,3 +44,10 @@ Feature: Data.gov.uk
     And I force a varnish cache miss
     When I search for "" in datasets
     Then I should see an accurate dataset count
+
+  @high
+  Scenario: Check that we don't get any s3 CSP errors for organogram previews
+    Given I am testing "https://data.gov.uk"
+    When I preview an organogram
+    Then I should see "View the full organogram"
+    And I don't get any s3 CSP errors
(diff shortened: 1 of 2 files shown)


features/data_gov_uk.feature
```diff
@@ -44,3 +44,10 @@ Feature: Data.gov.uk
     And I force a varnish cache miss
     When I search for "" in datasets
     Then I should see an accurate dataset count
+
+  @high
+  Scenario: Check that we don't get any s3 CSP errors for organograms previews
+    Given I am testing "https://data.gov.uk"
+    When I preview an organogram
+    Then I can see the 'View the full organogram' button
```
thomasleese: Here you can use `Then I should see "View the full organogram"` and you don't need the custom `Then` entry for this.

features/step_definitions/datagovuk_steps.rb
```diff
@@ -1,3 +1,5 @@
+require 'pry'
```
thomasleese: I think this was left in from debugging - do we still need this?

kentsanggds: will remove!

PR #533, opened 2019-05-20
Add finder frontend XSS tests

koetsier:
Add XSS tests for finder frontend.

Attempts a number of the "naughty strings",
https://github.com/minimaxir/big-list-of-naughty-strings to see
if finder-frontend can be forced to run javascript to display
an alert.

This tests inputting text in the keywords text field through javascript 
and by submitting the keywords through the URL

Trello: https://trello.com/c/eaa0vKa2/726-add-tests-for-xss-in-text-fields-in-finder-frontend-and-smokey-l

### DIFF
--- a/features/finder_frontend.feature
+++ b/features/finder_frontend.feature
@@ -53,6 +53,59 @@ Feature: Finder Frontend
     Then I should see filtered documents
 
   @high
-  Scenario: Check malicious code does not execute
-    When I visit "/government/organisations/hm-revenue-customs/contact?keywords=<script>alert(document.cookie);</script>"
-    Then I should see "&lt;script&gt;alert(document.cookie)"
+  Scenario Outline: Check malicious code does not execute
+    When I visit the "<finder>" finder with keywords <keyword>
+    Then There should be no alert
+    When I visit the "<finder>" finder without keywords
+    And I fill in the keyword field with <keyword>
+    Then There should be no alert
+    And I should see the string <keyword>
+
+  Examples: news-and-communications finder
+    | keyword                                              |  finder                      |
+    |<script>alert(123)</script>                           | news-and-communications      |
+    |&lt;script&gt;alert(&#39;123&#39;);&lt;/script&gt;    | news-and-communications      |
+    |<img src=x onerror=alert(123) />                      | news-and-communications      |
+    |<svg><script>123<1>alert(123)</script>                | news-and-communications      |
+    |\"><script>alert(123)</script>                        | news-and-communications      |
+    |'><script>alert(123)</script>                         | news-and-communications      |
+    |><script>alert(123)</script>                          | news-and-communications      |
+    |</script><script>alert(123)</script>                  | news-and-communications      |
+    |< / script >< script >alert(123)< / script >          | news-and-communications      |
+    | onfocus=JaVaSCript:alert(123) autofocus              | news-and-communications      |
+    |\" onfocus=JaVaSCript:alert(123) autofocus            | news-and-communications      |
+    |' onfocus=JaVaSCript:alert(123) autofocus             | news-and-communications      |
+    |＜script＞alert(123)＜/script＞                        | news-and-communications      |
+    |<sc<script>ript>alert(123)</sc</script>ript>          | news-and-communications      |
+    |--><script>alert(123)</script>                        | news-and-communications      |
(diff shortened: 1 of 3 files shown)


features/step_definitions/smokey_steps.rb
```diff
@@ -57,6 +57,28 @@
...
+end
+
+And /^I should see the string (.+)$/ do |content|
+  expect(page.find_field('keywords').value).to eq(content)
+end
+
+And /^I fill in the keyword field with (.+)$/ do |content|
+  page.fill_in 'keywords', with: content
+  page.find_field('keywords').native.send_keys(:return)
+  page.find('button[data-name="keywords"]', match: :first)
```
bilbof: Does this submit the keywords?

features/finder_frontend.feature
```diff
@@ -53,6 +53,58 @@ Feature: Finder Frontend
...
   @high
-  Scenario: Check malicious code does not execute
-    When I visit "/government/organisations/hm-revenue-customs/contact?keywords=<script>alert(document.cookie);</script>"
-    Then I should see "&lt;script&gt;alert(document.cookie)"
+  Scenario Outline: Check malicious code does not execute
+    When I visit the "<finder>" finder with keywords <keyword>
+    Then There should be no alert
+    And I fill in the keyword field with <keyword>
+    Then There should be no alert
+    And I should see the string <keyword>
```
bilbof: I suppose JS needs to be enabled for this test?

features/step_definitions/smokey_steps.rb
koetsier: line 77 does. This just waits for the tag to appear so we know the page has loaded

features/step_definitions/smokey_steps.rb
```diff
@@ -57,6 +57,28 @@
...
 
+When /^I visit the "([^"]+)" finder with keywords (.*)$/ do |finder, keywords|
+  path = "/search/#{finder}?keywords=#{keywords}"
+  visit_path path
+end
+
+And /^There should be no alert$/ do
+  expect {
+    page.driver.browser.switch_to.alert.accept
+  }.to raise_error(Selenium::WebDriver::Error::NoSuchAlertError)
```
bilbof: This seems to try to close an alert and then raise an error if there isn't an alert to close. Is that right?

Is there another way to check if an alert has been created?

features/finder_frontend.feature
koetsier: I assume all these tests run with JS enabled

features/step_definitions/smokey_steps.rb
koetsier: Not that I could find.

features/finder_frontend.feature
bilbof: Do you need the `@javascript` flag as in `/features/finders.feature`?

features/step_definitions/smokey_steps.rb
bilbof: Is it possible to do something like this? Like in features/step_definitions/filtering_steps.rb

```
And(/^I fill in some keywords$/) do
  fill_in 'Search', with: "Keyword1 Keyword2\n"
end
```

features/finder_frontend.feature
koetsier: This appears not to be necessary.

bilbof: Will filling in the keyword field have an effect, given that queries are cached when JS is enabled?

koetsier: good point. I added a reload so the cache should be emptied

features/step_definitions/smokey_steps.rb
koetsier: fixed

bilbof: Does `"#{content}"` escape the text before it's added?

barrucadu: As far as I can tell, this will emulate a human typing in that text.  So the value of the field will be the unescaped text.

kevindew:
@koetsier I'm seeing these break quite regularly for Smokey (I'm on 2nd line) - I'm not quite sure why they're failing.

It does look like there are a heck of a lot of tests here for Smokey could these not be served better in the individual application? It feels like we just need a single cursory in Smokey rather than an exhaustive suite.

PR #537, opened 2019-05-24
Reduce quantity of finder frontend XSS tests and address flakiness

kevindew:
This is raised to address problems we have been seeing from 2nd line regarding the recently added Finder Frontend XSS tests: https://github.com/alphagov/smokey/pull/533

This does some work to reduce the time these tests take to run by about 4-5 seconds per test, it adds a greater timeout to try stop the tests failing as often as they do, and it reduces the amount of tests as these seem too thorough for something testing at an end-to-end level.

Further details in commit messages.

### DIFF
--- a/features/finder_frontend.feature
+++ b/features/finder_frontend.feature
@@ -61,51 +61,7 @@ Feature: Finder Frontend
     Then There should be no alert
     And I should see the string <keyword>
 
-  Examples: news-and-communications finder
-    | keyword                                              |  finder                      |
-    |<script>alert(123)</script>                           | news-and-communications      |
-    |&lt;script&gt;alert(&#39;123&#39;);&lt;/script&gt;    | news-and-communications      |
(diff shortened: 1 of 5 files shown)


kevindew:
> Stellar job! 👏 great commit messages.

Thanks 👍 

> Can we ensure the removed XSS tests are moved to the finder repo itself?

Are you on the team that looks after this? I can't see the trello card of the original one https://trello.com/c/eaa0vKa2/726-add-tests-for-xss-in-text-fields-in-finder-frontend-and-smokey-l but it'd make sense to pull that card back from done if there are now gaps in the criteria.

I would question the value of all of those tests myself though, essentially we're testing that a string has it's HTML entities encoded 21 times. I'd have thought a single test that checked we are encoding the HTML entities would suffice

ChrisBAshton:
I can't see the Trello card either. Happy to leave on the basis that the tests provided little value beyond the first one.

PR #542, opened 2019-05-29
Use the "wait-until" logic for search analytics checks

barrucadu:
Checking for the async analytics requests is pretty flaky.  Moving them to the end, after all the sync requests, hasn't helped.  So instead use the `wait_until` function from the government-frontend step definitions to wait until the request has been made.  There is a max delay of 5s.

---

[Trello card](https://trello.com/c/u7D6Yx0c/770-fix-flaky-search-analytics-test)

### DIFF
--- a/features/step_definitions/search_steps.rb
+++ b/features/step_definitions/search_steps.rb
@@ -6,11 +6,11 @@ When /^I expand the search options$/ do
   click_link "+ Show more search options"
 end
 
-And /^I go to the next page$/ do
+When /^I go to the next page$/ do
   click_link "Next page"
 end
 
-And /^I click result (.*)$/ do |n|
+When /^I click result (.*)$/ do |n|
   result = page.all(".finder-results li a")[n.to_i - 1]
   click_link result.text
 end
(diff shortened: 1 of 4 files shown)


features/step_definitions/search_steps.rb
```diff
@@ -34,28 +34,27 @@
...
 
 Then /^search analytics for "(.*)" are reported$/ do |term|
-  found = false
   sought = "dp=#{CGI::escape("/search/all?keywords=#{term.sub(' ', '+')}")}"
-  $proxy.har.entries.each do |e|
-    found = true if e.request.url.include? sought
-  end
-  expect(found).to be(true)
+  wait_until { proxy_has_request_containing sought }
+  expect(proxy_has_request_containing sought).to be(true)
```
bilbof: This will raise a timeout error rather than explaining what happened. Would it be useful to rescue the error and raise an expectation type error so when it fails it'll be clearer why?

barrucadu: I think it's fine, because the exception will point to the `wait_until` line which has the condition.

PR #551, opened 2019-06-17
Test backend validation on email field

leenagupte:
This test was failing because it was failing the browser validation on input fields
of type="email". This step removes the browser validation from the email field, forcing
the backend validation to run.


features/step_definitions/feedback_steps.rb
```diff
@@ -1,4 +1,5 @@
 When /^I input malicious code in the (.*) field$/ do |field|
+  page.execute_script('$("form").attr("novalidate", "novalidate")')
```
tijmenb: Maybe we could add a comment why this is necessary?

PR #552, opened 2019-06-18
Add smoke tests for cache

bilbof:
This adds a few small checks that caching works

It checks only the homepage, and that post requests
aren't cached.

At the moment CDN changes aren't checked specifically
by the smoke tests.

Trello: https://trello.com/c/IPbLTDLq/806

### DIFF
--- a/features/caching.feature
+++ b/features/caching.feature
@@ -1,0 +1,23 @@
+@aws
+Feature: Caching
+  Tests that pages are correctly cached
+
+  Background:
+    Given I am testing through the full stack
+
+  @normal
+  Scenario: Check council lookup
+    When I try to post to "/find-local-council" with "postcode=WC2B+6SE" without following redirects
+    Then I should not hit the cache
+    Then I should see "camden"
+
+  @normal
+  Scenario: Check licence lookup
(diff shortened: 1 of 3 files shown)


features/caching.feature
```diff
@@ -0,0 +1,23 @@
...
+  @normal
+  Scenario: Check licence lookup
+    When I try to post to "/busking-licence" with "postcode=E20+2ST" without following redirects
+    Then I should not hit the cache
+    Then I should see "newham"
+
+  @normal
+  Scenario: Check homepage is served by Varnish
+    When I request "/"
+    Then I should hit the cache
```
barrucadu: This has a chance of failing when the homepage is updated

bilbof: Yeah. I thought about setting it to `low` priority, given it has a chance of failing, albeit low.

We can't specify which edge node we'll hit on each request, so I couldn't find a way to force the page to be cached before we test for it. What do you think?

barrucadu: I think set it to `low`

bilbof: Sure thing, I amended the commit.

PR #566, opened 2019-07-17
Fix issues with WAF feature

thomasleese:
The WAF feature is currently failing with "undefined" errors:

```
1 scenario (1 undefined)
3 steps (2 undefined, 1 passed)
0m20.382s

You can implement step definitions for undefined steps with these snippets:

When("I send a GET request to \/") do
  pending # Write code here that turns the phrase above into concrete actions
end

Then("the response status should be {int}") do |int|
  pending # Write code here that turns the phrase above into concrete actions
end
```

This should fix the undefined errors.

### DIFF
--- a/features/step_definitions/waf_steps.rb
+++ b/features/step_definitions/waf_steps.rb
@@ -1,11 +1,10 @@
-
 Given /^I set header ([^\s]+) to ([^\s]+)/ do |header_name, header_value|
-	@headers ||= {}
-	@headers[header_name.to_sym] = header_value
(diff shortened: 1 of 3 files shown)


thomasleese:
Yeah, all the `Then`s (and `Given`s etc) are automatically available to all the tests.

The status code one is defined here: https://github.com/alphagov/smokey/blob/dba87a0767b8a04331e1fbfbe55eef0f640a24b4/features/step_definitions/smokey_steps.rb#L159-L172

PR #567, opened 2019-07-17
Force a cache miss when testing WAF

thomasleese:


chrisfarms:
Does a "varnish cache miss" imply a CDN cache miss too?

thomasleese:
Yeah... it's not very clever, all it does is add a cache busting string to the end of the URL.

PR #578, opened 2019-08-02
Add scenario for when a page is not useful

benthorner:
https://github.com/alphagov/govuk_publishing_components/issues/997

This adds a scenario to check a user is able to successfully submit a
request to give feedback when they don't find a page is useful. While
this feature exists on most pages on GOV.UK, this check only visits the
homepage to superficially cover the behaviour.

Co-Author: @bilbof

### DIFF
--- a/features/feedback.feature
+++ b/features/feedback.feature
@@ -19,3 +19,11 @@ Feature: Feedback
   Scenario: Check the FoI page loads correctly
     When I visit "/contact/foi"
     Then I should see "How to make a freedom of information (FOI) request"
+
+  @normal
+  Scenario: Check "is this page useful?" email survey
+    When I visit "/"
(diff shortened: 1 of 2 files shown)


features/support/env.rb
```diff
@@ -57,13 +57,7 @@
 # Licensify admin doesn't have favicon.ico so block requests to prevent errors
 proxy.blacklist(/^https:\/\/licensify-admin(.*)\.publishing\.service\.gov\.uk\/favicon\.ico$/i, 200)
 
-chromedriver_from_path = File.which("chromedriver")
```
bilbof: Heads up we should keep this stuff, though we needed to remove it for local development.

benthorner: Oops - removed :-).

PR #579, opened 2019-08-02
Fix feature tests for draft pages

chao-xian:
The previous commit changed the candidate page for draft stack
but didn't include updating the spec file itself and is pointing
to a newly published page that does not yet exist on integration.


features/step_definitions/draft_environment_steps.rb
```diff
@@ -25,7 +25,7 @@
 end
 
 Then /^I should be on the case study page$/ do
-  expect(page.current_path).to eq("/government/case-studies/loopwheels-delivering-a-smoother-ride-for-wheelchair-users")
+  expect(page.current_path).to eq("//government/case-studies/example-case-studies-eu-citizens-rights-in-the-uk")
```
thomasleese:
```suggestion
  expect(page.current_path).to eq("/government/case-studies/example-case-studies-eu-citizens-rights-in-the-uk")
```

chao-xian: 🤦‍♂

PR #582, opened 2019-08-05
Add basic auth headers to DGU test

kentsanggds:
## What

Smokey tests for DGU are failing on Integration as the Integration website needs basic auth, so adding the basic auth headers should fix them.


kentsanggds:
Tried it on Integration where Smokey is failing and it passed

PR #585, opened 2019-08-09
Add youtube-nocookie to the proxy blacklist

thomasleese:
I think this should solve the YouTube problem we were seeing before as all requests to that address should result in a 200 successful response.

This allows us to revert 4480e78 and go back to using the old case study which had an embedded YouTube video.

### DIFF
--- a/features/draft_environment.feature
+++ b/features/draft_environment.feature
@@ -15,7 +15,7 @@ Feature: Draft environment
   @normal @draft
   Scenario: Check visiting a page served by government-frontend
     When I try to login as a user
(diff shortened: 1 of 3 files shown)


thomasleese:
The other case study embeds a YouTube video. I'm not sure if that was chosen intentionally, but I think it's useful to include an example which does so in the test.

cbaines:
> The other case study embeds a YouTube video. I'm not sure if that was chosen intentionally, but I think it's useful to include an example which does so in the test.

Ok, fair enough.

PR #595, opened 2019-09-09
Add smoke tests for Brexit checker

theseanything:
This adds basic tests the Brexit check rendered by finder-frontend.

### DIFF
--- a/features/brexit_check.feature
+++ b/features/brexit_check.feature
@@ -1,0 +1,24 @@
+Feature: Get Ready for Brexit Check
+
+  Background:
+    Given I am testing through the full stack
+    And I force a varnish cache miss
+
+  @high
+  Scenario: Brexit checker shows results
+    When I start the checker
+    And I answer the nationality question
+    And I answer the where do you live question
+    And I skip all other questions
+    Then I should see the results page
+    And I should see confirmation of my answers
+    And I should see answers applicable to me
+
+  @high
+  Scenario: Check that the Brexit checker can be subscribed to
+    When I visit "/get-ready-brexit-check/questions"
+    And I answer the nationality question
+    And I skip all other questions
+    And I click Subscribe link
+    And I click Subscribe button
+    Then I should enter the email subscription workflow
--- a/features/step_definitions/brexit_check_steps.rb
+++ b/features/step_definitions/brexit_check_steps.rb
@@ -1,0 +1,46 @@
+When "I start the checker" do
+  visit_path "/get-ready-brexit-check"
+  click_link "Start now"
+end
+
+When "I answer the nationality question" do
+  choose('British', visible: false)
+  click_button "Next"
+end
+
+When "I answer the where do you live question" do
+  choose('Another EU country', visible: false)
+  click_button "Next"
+end
+
+When "I skip all other questions" do
+  while page.has_current_path?('/get-ready-brexit-check/questions', ignore_query: true)
+    click_button "Next"
+  end
+end
+
+Then "I should see the results page" do
+  expect(page).to have_content("What you need to do now to prepare for Brexit")
+end
+
+Then "I should see confirmation of my answers" do
+  expect(page).to have_content("You are a British national")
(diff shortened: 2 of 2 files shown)


features/brexit_check.feature
```diff
@@ -0,0 +1,26 @@
+Feature: Get Ready for Brexit Check
+
+  Background:
+    Given I am testing through the full stack
+    And I force a varnish cache miss
+
+  @high
+  Scenario: Brexit checker shows results
+    Given I am a UK national
```
benthorner: This was surprising, as it seems to exist purely for the purpose of a subsequent step. I had to read as I couldn't guess what 'being a UK national' translates to on GOV.UK. Could we just assume the step that answers the question is going to inject some sensible data?

theseanything:
> * I wonder if we can identify a 'most important flow' to test here? Doesn't seem like a critical thing, and there's definitely a compromise with making this test robust to question/action changes. I'm just wondering if we're considering any further iterations on this.

Thanks for the comments! Yup - agree, for now I've assumed that if we can answer a few questions and get results - we can rely on app feature tests to make sure more critical paths or question combinations work. This is a tradeoff to make the smokey test less coupled the structure of the questions, so allowing us to make change more easily. But we could defo iterate this!

> Would this feature be more appropriate in `features/finder_frontend.feature`? If it's going on it's own, perhaps it's worth extending it to include the whole landing page journey, to more explicitly identify it as a non-app-specific scenario set.

Keeping this as a separate will help isolate this code and (tech debt) if we need to remove it in the future. Not sure the landing page is seen as part of the Brexit checklist journey, as we have links from multiple places, not just the landing page, to the checklist "start page".

features/brexit_check.feature
theseanything: Updated to remove these given clauses.

features/step_definitions/brexit_check_steps.rb
```diff
@@ -0,0 +1,60 @@
+# Start page
+# ==========
+
+When "I start the checker" do
+  visit_path "/get-ready-brexit-check"
+  click_link "Start now"
+  visit_path "/get-ready-brexit-check/questions" # cache busts
```
benthorner: Do we still want to get rid of this?

features/step_definitions/brexit_check_steps.rb
```diff
@@ -0,0 +1,60 @@
+# Start page
```
benthorner: Minor, but I don't think these comments add any value.

theseanything:
@benthorner should be ready for review again

features/step_definitions/brexit_check_steps.rb
```diff
@@ -0,0 +1,47 @@
...
+end
+
+Then "I should see confirmation of my answers" do
+  expect(page).to have_content("What you need to do now to prepare for Brexit")
+  expect(page).to have_content("You are a British national")
+  expect(page).to have_content("You live in the EU")
+end
+
+Then "I should see answers tailored to me" do
+  expect(page).to have_content("Apply to be a resident in the EU country you live in")
```
benthorner: Minor, but I wonder if the description of this step suggests we should check that other actions aren't visible. What do you think?

features/step_definitions/brexit_check_steps.rb
```diff
@@ -0,0 +1,47 @@
+When "I start the checker" do
+  visit_path "/get-ready-brexit-check"
+  click_link "Start now"
+end
+
+When "I answer the nationality question" do
+  choose('British', visible: false)
```
benthorner: Minor, but I don't think we need the `visible: false` modifiers for these non-invisible elements.

theseanything: Need as the radio buttons are hidden to support custom styling.

features/step_definitions/brexit_check_steps.rb
theseanything: Modified the description. 👍

PR #608, opened 2019-10-02
Fix error message selector

thomasleese:
This was changed in https://github.com/alphagov/smart-answers/pull/4132.

### DIFF
--- a/features/step_definitions/smartanswers_steps.rb
+++ b/features/step_definitions/smartanswers_steps.rb
@@ -10,7 +10,7 @@ end
 Then /^the slug should be (valid|invalid)$/ do |valid_or_invalid|
   is_valid = valid_or_invalid == "valid"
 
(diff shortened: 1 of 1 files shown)


features/step_definitions/smartanswers_steps.rb
```diff
@@ -10,7 +10,7 @@
 Then /^the slug should be (valid|invalid)$/ do |valid_or_invalid|
   is_valid = valid_or_invalid == "valid"
 
-  error_message = "#current-error"
+  error_message = "#govuk-error-message"
```
vanitabarrett: I think you want this to be a class rather than an ID? `.govuk-error-message`

thomasleese: Yes... good point! I've amended the commit.

PR #622, opened 2019-12-04
Remove check for email frequency preselection

vanitabarrett:
Trello: https://trello.com/c/KYrvoZmz/306-double-opt-in-stop-preselecting-the-frequency-when-subscribing
Relates to: https://github.com/alphagov/email-alert-frontend/pull/624/commits

In https://github.com/alphagov/email-alert-frontend/pull/624/commits we are stopping the preselection of email frequency when users sign up for new subscriptions. We can therefore remove the tests associated with this functionality.


benthorner:
Curious this has no build 🤔

PR #657, opened 2020-03-31
Correct Smokey tests to match new signup flow

vanitabarrett:
We have made changes to the [email signup flow](https://github.com/alphagov/email-alert-frontend/pull/722) and the [content on those pages](https://github.com/alphagov/email-alert-frontend/pull/732).


laurentqro:
thanks @vanitabarrett

PR #669, opened 2020-06-01
Fix sign up link text

thomasleese:
This is similar to https://github.com/alphagov/smokey/pull/659 but for a transport taxon. I'm not sure why we've only had to change it now.

https://www.gov.uk/email-signup?topic=/topic/transport/motorways-major-roads is the page we get to with the "Sign up" button.


vanitabarrett:
Sorry both, should have spotted this! 

This is actually an old topic (as opposed to the new taxonomy) - @chao-xian recently [switched topics to use email-alert-frontend for the start of the sign-up journey rather than a custom view in Collections](https://github.com/alphagov/collections/pull/1749)

PR #678, opened 2020-06-19
Add a feature for the Publishing API

benthorner:
https://trello.com/c/aW59k3Tf/170-check-that-publishing-api-has-at-least-one-smoke-test-that-ensures-its-actually-running-and-tag-those-tests-with-app-appname


benthorner:
Closing, as these tests aren't easy to add, and don't clearly represent a coverage issue.

PR #681, opened 2020-06-23
Add and tag features for the Publishing API

benthorner:

### DIFF
--- a/features/publishing_tools.feature
+++ b/features/publishing_tools.feature
@@ -1,3 +1,4 @@
+@app-publishing-api
 Feature: Publishing Tools
 
   @high
--- a/features/publishing_api.feature
+++ b/features/publishing_api.feature
@@ -1,0 +1,6 @@
+@app-publishing-api @local-network
+Feature: Publishing API
+  Scenario: Healthcheck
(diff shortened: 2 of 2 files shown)


features/publishing_tools.feature
```diff
@@ -1,3 +1,4 @@
+@app-publishing-api
```
ChrisBAshton, line 1: What does this tagging actually achieve? Do we ever just run a subset of smokey tests by its tags, or does it produce more useful error output, or is this purely for categorisation / local development?

benthorner: Indeed we do: https://github.com/alphagov/govuk-puppet/pull/10387

ChrisBAshton: Excellent :) Will we have any corresponding docs on https://docs.publishing.service.gov.uk/?

benthorner: Yeah, that's on my list to do. I'd like to get this in if possible, and address the documentation separately.

I've made a start on that [here](https://github.com/alphagov/smokey/pull/683), if you'd like to review that 😃.

PR #688, opened 2020-06-26
Fix non-matching step for Publishing API

benthorner:


benthorner:
Tested and works now 😄

PR #691, opened 2020-06-29
Rewrite and extend docs for adding/changing tests

benthorner:
https://trello.com/c/aW59k3Tf/170-activate-continuous-deployment-for-the-publishing-api

Related to: https://github.com/alphagov/smokey/pull/681#discussion_r444320128

This PR does two things:

- Rewrite doc about Smokey and Icinga
- Document convention for "@app-" tags

Please see the commits for more details.

PR #693, opened 2020-06-30
Treat all scenarios with the same priority

benthorner:

**Deadline for discussion: 2020-07-08**

Related to: https://github.com/alphagov/smokey/pull/691

This changes the alerting behaviour for the results of the
Smokey loop so that are scenarios are considered 'normal'
priority, due to the lack of any consistent definition of
'high' and 'low' priorities, and the complete absence of
'urgent'. Please see the commits for more details.

Follow-up tasks:

- [ ] Remove [redundant code to support priorities](https://github.com/alphagov/smokey/blob/master/nagios_check_cache.py)
- [ ] Remove [redundant priority-based checks in Puppet](https://github.com/alphagov/govuk-puppet/blob/4ec2949332079570be5493753bb84a983218a444/modules/icinga/manifests/check_feature.pp)

## Before 🤷‍♂️ 

Screenshot 2020-06-30 at 10 56 11

## After 🤷‍♂️ 

Screenshot 2020-06-30 at 10 44 35

### DIFF
--- a/nagios_check_cache.py
+++ b/nagios_check_cache.py
@@ -92,13 +92,11 @@ class FeatureTestRun(object):
         result_details.add("\n")
 
     def set_status_for_scenarios_in_feature(self, feature, priority, result_details):
-        if 'elements' not in feature:
-            log_result_and_exit(0, "OK: Feature %s has no steps at any priority" % feature['id'])
+        if priority != '@normal':
+            return
+
         for scenario in feature['elements']:
-            if 'tags' in scenario:
-                for tag in scenario['tags']:
-                    if tag['name'] == priority:
-                        self.log_details_and_set_status_for_scenario(result_details, feature, priority, scenario)
+            self.log_details_and_set_status_for_scenario(result_details, feature, priority, scenario)
 
     def set_status_for_feature(self, feature_name, json_data, priority):
         feature_uri = 'features/' + feature_name + '.feature'
--- a/docs/writing-tests.md
+++ b/docs/writing-tests.md
@@ -2,14 +2,7 @@
 
 ## Alerting in Icinga
 
-In order to alert on a failing scenario in Icinga, you need to:
-
-- Ensure the feature is [listed in the Puppet config](https://github.com/alphagov/govuk-puppet/blob/d7af16e96aed682facb5cf5bc3e3972510c64ca2/hieradata_aws/integration.yaml#L378)
-
-- Tag the scenario with a [priority "@<tag>"](https://github.com/alphagov/govuk-puppet/blob/4ec2949332079570be5493753bb84a983218a444/modules/icinga/manifests/check_feature.pp)
-
-  - `@urgent` tags will trigger will [trigger out-of-hours callouts](https://github.com/alphagov/govuk-puppet/blob/51d7f7b648e8fde5aae3adfb774ec3bca6325bd8/modules/monitoring/manifests/contacts.pp#L163)
-  - Other tags are purely to help us prioritise fixing things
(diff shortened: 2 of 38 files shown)


nagios_check_cache.py
```diff
@@ -92,13 +92,11 @@ def log_details_and_set_status_for_scenario(self, result_details, feature, prior
         result_details.add("\n")
 
     def set_status_for_scenarios_in_feature(self, feature, priority, result_details):
-        if 'elements' not in feature:
-            log_result_and_exit(0, "OK: Feature %s has no steps at any priority" % feature['id'])
+        if priority != '@normal':
```
ChrisBAshton, line 95: Could you explain a bit more in the commit message, e.g. how 'priority' is defined now (I assume it defaults to `@normal` somewhere?)

benthorner: I can explain more, but I'll need to clarify what to explain 😄. There's no "default" at the moment: anything without one of these tags won't get picked up by Icinga, was is surprising to me. I could add that in?

benthorner:
> I think you could do Remove redundant code to support priorities as part of this PR, unless there's a good reason not to?

The reason not to is because we can't _completely_ remove it here, until [the corresponding Puppet bit](https://github.com/alphagov/govuk-puppet/blob/4ec2949332079570be5493753bb84a983218a444/modules/icinga/manifests/check_feature.pp) that's using it is gone. We should be able to get rid of *all* mention of "priority", if we do it in stages.

nagios_check_cache.py
ChrisBAshton: I'm still confused - if we've removed '@normal' from all the features, why is this line here?

Yes, worth clarifying if a tag is needed in order to be picked up by Icinga :)

> The reason not to is because we can't _completely_ remove it here, until [the corresponding Puppet bit](https://github.com/alphagov/govuk-puppet/blob/4ec2949332079570be5493753bb84a983218a444/modules/icinga/manifests/check_feature.pp) that's using it is gone. We should be able to get rid of *all* mention of "priority", if we do it in stages.

Can we not remove the puppet bit first?

benthorner: I'll try and explain: this script is invoked **by Puppet** with a specific priority. Previously we would compare the specified priority with the tags on each scenario (in the JSON test output), and only report on one that matched. Cucumber has no internal concept of "priority" or "default tag", so any feature without one of our priority tags wasn't picked up by this script and alerted on. This is unnecessarily surprising.

The change here is to remove the part of the script that matches against the priority tag. Puppet is still going to call the script multiple times with different priorities, and the change here means we'll only report back for the time it calls it with the "normal" priority (which the script appends an `@` to).

> Can we not remove the puppet bit first?

We can't remove the Puppet bit because it's priority-based (see [here](https://github.com/alphagov/govuk-puppet/blob/4ec2949332079570be5493753bb84a983218a444/modules/icinga/manifests/check_feature.pp). We need to start ignoring the priority here first (i.e. make the concept redundant) before we can remove it in Puppet.

Does that help?

ChrisBAshton: That makes sense, thanks. Can you add that second paragraph to the commit message?

benthorner: Sure, I've rephrased and updated the last commit. Thanks for the questions - I now realise the change is quite difficult to understand unless you're familiar with the script and how/where it's used!

ChrisBAshton: Thanks - message looks good except there is a `[1]` that suggests there's a URL missing?

benthorner: Oops - updated 😸

1pretz1:
Looks good 💯 

The priorities don’t mean a great deal to me personally like I wouldn’t know what tests would be classed as what priority without trawling through the test suite so I don't see an issue with standardising all the features as the same priority! 👍

PR #695, opened 2020-07-01
Add a PR template to encourage manual testing

benthorner:

Tripped me up on: https://github.com/alphagov/smokey/pull/692

## Testing

> This app doesn't have CI setup, and it would be misleading to do so:
> the behaviour of the tests changes depending on the environment they
> are run in.

## Deployment

Steps to deploy a change:

> The manual [Smokey job][] will pick up changes on `master` automatically. 
> You only need to do a manual deployment for the [continuous Smokey loop][].

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,29 @@
+## Testing
+
+**You should manually test your PR before merging.**
+
+This app doesn't have CI setup, and it would be misleading to do so:
+the behaviour of the tests changes depending on the environment they
+are run in. You should manually test in applicable environments,
+until you are confident your change is not a breaking one.
+
+Example steps to test in Integration:
+
+- Click "Configure" for the [Smokey job][]
+- Change the "Branch specifier" to the name of your branch
+- Run a new build of the Smokey project and check the results
+
+## Deployment
+
+**You need to manually deploy your PR after merging.**
+
+Steps to deploy a change:
+
+- Run the [Smokey deploy job][] to deploy the [continuous Smokey loop][]
+- Repeat this in Staging and Production
+
+> The manual [Smokey job][] will pick up changes on `master` automatically. You only need to do a manual deployment for the [continuous Smokey loop][].
+
+[Smokey job]: https://deploy.integration.publishing.service.gov.uk/job/Smokey/
+[continuous Smokey loop]: https://github.com/alphagov/govuk-puppet/blob/master/modules/monitoring/templates/smokey-loop.conf
+[Smokey deploy job]: https://deploy.integration.publishing.service.gov.uk/job/Smokey_Deploy/
--- a/README.md
+++ b/README.md
@@ -62,3 +62,4 @@ This repo also contains several scripts to support external systems running the 
 
 - [troubleshooting](docs/troubleshooting.md)
 - [writing-tests](docs/writing-tests.md)
+- [pr-template](.github/pull_request_template.md)


.github/pull_request_template.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Deployment
+
+**You need to manually deploy your PR after merging.**
+
+- [ ] Ran `Smokey deploy` job in all environments
+
+This app is deployed in two ways (see beginning of [the README](README.md)):
+
+- The [Smokey job][] will pick up the changes on `master`.
```
theseanything, line 23: This could more explicit that there is nothing to do.
```suggestion
- The [Smokey job][] will pick up the changes on `master` automatically, no further action required.
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,26 @@
...
+- The [continuous Smokey loop](https://github.com/alphagov/govuk-puppet/blob/master/modules/monitoring/templates/smokey-loop.conf) needs manually deploying
```
theseanything, line 24: Should we add instructions or a link to how to deploy? Is it the normal process?

theseanything: Or is this covered by the 'deploy smokey' job?

.github/pull_request_template.md
```diff
@@ -0,0 +1,26 @@
+## Testing
+
+This app doesn't have CI setup, and it would be misleading to do so:
+the behaviour of the tests changes depending on the environment they
+are run in. **You should manually test your PR before merging.**
+
+- [ ] Tested change in at least one environment
```
theseanything, line 7: Nit: To make it easier to read, you can start with the instructions of what the user needs to do, followed by explanation.

```suggestion
**You should manually test your PR before merging.**

- [ ] Tested change in at least one environment

This app doesn't have CI setup, and it would be misleading to do so:
the behaviour of the tests changes depending on the environment they
are run in. 
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,26 @@
...
+>
+> - Click "Configure" for the [Smokey job][]
+> - Change the "Branch specifier" to the name of your branch
+> - Run a new build of the Smokey project and check the results
+
+## Deployment
+
+**You need to manually deploy your PR after merging.**
+
+- [ ] Ran `Smokey deploy` job in all environments
```
theseanything, line 19: Does this need to be a checkbox? Just thinking the PR will be merged at this point - and some people may not fill it out. Which could be confusing for people looking at the PRs in future.

.github/pull_request_template.md
benthorner: I've rewritten this to explain the steps.

.github/pull_request_template.md
benthorner: Sure, I've rewritten this bit.

.github/pull_request_template.md
benthorner: Yeah, they were encouraging but not necessary for me. I've removed them.

benthorner:
@theseanything thanks for your comments. I've rewritten this to incorporate them, and updated the description of the PR with the new version. Can you take another look?

.github/pull_request_template.md
```diff
@@ -0,0 +1,28 @@
+## Testing
+
+**You should manually test your PR before merging.**
+
+Example steps to test in Integration:
```
theseanything, line 5: Are we asking people to test in all environments? or just integration? or any environment they want?

benthorner: It's a good question. I didn't want to be too prescriptive, as _theoretically_ people should follow our normal deployment process and therefore do a branch deploy of their change in each environment (since a merge "deploys" to all simultaneously). Do you think people won't exercise common sense here?

theseanything: Hmm, yeah its tricky. I guess we get into the grey area of common sense - I'd probably hesitate to test in production. I think if people **_should_** be testing in every environment, then we should be explicit and let people use common sense if they deviate from that?

benthorner: @theseanything I've updated this as-per our discussion. Can you take another look?

PR #697, opened 2020-07-01
Update the brexit checker test

sihugh:
The start page has now been redirected to the /transition page now, so we should start from there instead.

### DIFF
--- a/features/step_definitions/brexit_check_steps.rb
+++ b/features/step_definitions/brexit_check_steps.rb
@@ -1,6 +1,6 @@
 When "I start the checker" do
-  visit_path "/get-ready-brexit-check"
-  click_link "Start now"
+  visit_path "/transition"
+  click_link "Check what's changing"
 end
 
 When "I answer the nationality question" do


features/step_definitions/brexit_check_steps.rb
```diff
@@ -1,6 +1,6 @@
 When "I start the checker" do
-  visit_path "/get-ready-brexit-check"
-  click_link "Start now"
+  visit_path "/transition"
+  click_link "Check what's changing"
```
1pretz1, line 3: Should this not be "Check what’s changing and sign up for email alerts" ?

sihugh: It could be, but I think the text is likely to be iterated by content people, so was hoping the test would match on a bit of the text so we maybe didn't have to tweak it as much.

1pretz1: That makes sense - sorry I didn't realise Capybara fuzzy matches by default which is cool -> https://rubydoc.info/github/jnicklas/capybara/#exactness.

sihugh: ikr!

PR #702, opened 2020-07-07
Remove path defaults for internal URLs

benthorner:
This is confusing and unnecessary for internal URLs. Ideally we
should drop the defaults for external URLs as well (because it's
so surprising), but that would require more extensive changes to
features that are relying on the defaults.


benthorner:
- [x] Tested with a full run of Smokey, and works!

PR #718, opened 2020-08-26
Specify finder-frontend tests to run against AWS

karlbaker02:
This PR specifies via the `@aws` tag to run the `@app-finder-frontend` tests against AWS, as that is where the app is served from. The reason for this is to fix the Smokey test for the finder-frontend healthcheck which is currently resolving against what would have been the public endpoint on Carrenza and which now no longer exists. This test has been consistently failing since it was introduced, and was not marked to test against AWS when other apps hosted on AWS were (see https://github.com/alphagov/smokey/commit/91a5ce94b7b96d09f5db87e63fbc3e400788de00).

## Testing

## Deployment

Steps to deploy a change:

### DIFF
--- a/features/finder_frontend.feature
+++ b/features/finder_frontend.feature
@@ -1,4 +1,4 @@
-@app-finder-frontend
+@aws @app-finder-frontend
 Feature: Finder Frontend
   These are pages that let you search within a set of similar looking documents.
 


issyl0 (approved):
As part of the cleanup of Carrenza, I'm looking forward to seeing the `@aws` tag gone and it just be the default. But I'm doubtful I'll still be here to see it.

PR #742, opened 2020-11-19
Fix not actually checking app healthchecks

benthorner:
https://trello.com/c/ILpSOFhq/201-enable-cd-for-quick-win-and-possible-quick-win-apps

Previously in scenarios requesting the /healthcheck endpoint we only
verified the response code was 200, but this will always be the case
even if the app is unhealthy. This adjusts the /healthcheck scenarios
to actually verify the endpoint reports an "OK" status.

## Testing

## Deployment

Steps to deploy a change:

### DIFF
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -71,4 +71,5 @@ Feature: Collections
   Scenario: Healthcheck
     Given I am testing "collections" internally
     When I request "/healthcheck"
-    Then I should get a 200 status code
+    Then JSON is returned
+    And I should see ""status":"ok""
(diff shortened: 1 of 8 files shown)


benthorner:
Need to fix a problem with the JSON endpoint for finder-frontend not being exposed.

benthorner:
Depends on outcome of: https://github.com/alphagov/govuk-puppet/pull/10852.

features/collections.feature
```diff
@@ -71,4 +71,5 @@ Feature: Collections
   Scenario: Healthcheck
     Given I am testing "collections" internally
     When I request "/healthcheck"
-    Then I should get a 200 status code
+    Then JSON is returned
+    And I should see ""status":"ok""
```
thomasleese, line 75: The double `""` is a little bit jarring to see, but I think it's fine.

thomasleese (approved):

benthorner:
Tested and works on Staging ✅. Just waiting for the prerequisite Puppet change ☝️ to roll out in Production.

PR #738, opened 2020-11-05
Add test for redirect from postcode checker to national guidance

JonathanHallam:
The postcode checker has been hidden to the public. This tests that the redirect to the appropriate guidance is in place.

## Testing

**You should manually test your PR before merging.**

This app doesn't have CI setup, and it would be misleading to do so:
the behaviour of the tests changes depending on the environment they
are run in. You should manually test in applicable environments,
until you are confident your change is not a breaking one.

Example steps to test in Integration:

- Click "Configure" for the [Smokey job][]
- Change the "Branch specifier" to the name of your branch
- Run a new build of the Smokey project and check the results

## Deployment

**You need to manually deploy your PR after merging.**

Steps to deploy a change:

- Run the [Smokey deploy job][] to deploy the [continuous Smokey loop][]
- Repeat this in Staging and Production

> The manual [Smokey job][] will pick up changes on `master` automatically. You only need to do a manual deployment for the [continuous Smokey loop][].

[Smokey job]: https://deploy.integration.publishing.service.gov.uk/job/Smokey/
[continuous Smokey loop]: https://github.com/alphagov/govuk-puppet/blob/master/modules/monitoring/templates/smokey-loop.conf
[Smokey deploy job]: https://deploy.integration.publishing.service.gov.uk/job/Smokey_Deploy/

PR #751, opened 2020-11-27
Remove tests for redirecting the postcode checker to national guidance

beccapearce:
These tests were added when the postcode checker was hidden to the
public and a redirect was put in place, since the postcode checker is
live again these tests are now failing.

Here is the original PR: https://github.com/alphagov/smokey/pull/738

This PR removes the tests since the redirect is no longer in place to get
smokey passing again.

## Testing

## Deployment

Steps to deploy a change:


barrucadu (approved):

leenagupte:
Could they be reinstated instead: https://github.com/alphagov/smokey/pull/748?

PR #756, opened 2020-12-08
Add Manuals Publisher healthcheck test

MuriloDalRi:
Dependent on https://github.com/alphagov/manuals-publisher/pull/1689

## Testing

## Deployment

Steps to deploy a change:

### DIFF
--- a/features/manuals_publisher.feature
+++ b/features/manuals_publisher.feature
@@ -1,0 +1,8 @@
+@app-manuals-publisher
+Feature: Manuals Publisher
+
+  @local-network
+  Scenario: Healthcheck
+    Given I am testing "manuals-publisher" internally
+    When I request "/healthcheck"
+    Then I should get a 200 status code


features/manuals_publisher.feature
```diff
@@ -0,0 +1,8 @@
+@app-manuals-publisher
+Feature: Manuals Publisher
+
+  @local-network
+  Scenario: Healthcheck
+    Given I am testing "manuals publisher" internally
+    When I request "/healthcheck"
+    Then I should get a 200 status code
```
ChrisBAshton, line 8: Looks like this needs a newline at the end of the file (I think that's the convention, though can't find a doc to back that up 🤔 )

features/manuals_publisher.feature
```diff
@@ -0,0 +1,8 @@
+@app-manuals-publisher
+Feature: Manuals Publisher
+
+  @local-network
+  Scenario: Healthcheck
+    Given I am testing "manuals publisher" internally
```
ChrisBAshton, line 6: I think this needs the hyphen (`manuals-publisher`), otherwise [Plek won't be able to find it](https://github.com/alphagov/smokey/blob/f5ff5b39b060fa7b933453c8b9644ec871c04e6e/features%2Fsupport%2Fbase_urls.rb#L24).

ChrisBAshton (commented):
Just a couple of small points:

features/manuals_publisher.feature
MuriloDalRi: Fixed

features/manuals_publisher.feature
MuriloDalRi: I think I fixed this as you were typing this comment haha

ChrisBAshton (approved):

PR #776, opened 2021-01-14
Re-enable malicious code test for finder-frontend

1pretz1:
Previously this test was disabled [1] over a month ago due to being
flakey. We believed that there was an issue with CORs and google
analytics which led to a card being created to investigate. Since then
I've tried re-enabling the test to replicate the flakeyness for
integration and staging which proved to be unsuccessful, as out of 30
runs, there were no failures. This commit re-enables the test as there
doesn't seem to be an issue anymore.

[1]: https://github.com/alphagov/smokey/pull/755

I've recorded all the builds as a comment on the [Trello card](https://trello.com/c/5hdexmoU/2292-investigate-fix-failing-smokey-test-for-finder-frontend-timebox-2-days).

## Testing

## Deployment

Steps to deploy a change:

### DIFF
--- a/features/finder_frontend.feature
+++ b/features/finder_frontend.feature
@@ -49,7 +49,6 @@ Feature: Finder Frontend
     Then I should see filtered documents
     And I should see an open facet titled "Case type" with non-blank values
 
-  @notintegration @notstaging @notproduction
   Scenario Outline: Check malicious code does not execute
     When I visit the "<finder>" finder with keywords <keyword>
     Then There should be no alert


ChrisBAshton (approved):
Weird - though we did suspect a change at Google's end, and presumably they've made another change.

Did you have a chance to think about whether there's a better approach to these tests? E.g. as they're equally applicable in all the frontend apps, not just finder-frontend.

1pretz1:
Thanks for the review @ChrisBAshton!

> though we did suspect a change at Google's end, and presumably they've made another change.

Yup, that's what I imagine happened as I couldn't see anything obvious from our end.

> Did you have a chance to think about whether there's a better approach to these tests? E.g. as they're equally applicable in all the frontend apps, not just finder-frontend.

I do agree that there should be some uniformity between frontend applications on how we test security. Had a ponder on what that may be and not quite sure what to suggest.

If following rails conventions it should be pretty difficult to open up these sort (XSS) of vulnerabilities. We also use Brakeman for most of our apps so you'd hope that simple security issues would be raised during test builds.

However to answer your question, for simplicity we could implement a basic test per app (smokey or via the app) that would run the same sort of test that's already used for [finder-frontend] and [feedback]. One drawback being is that the current check is quite primitive and only tests one sort of attack and if an app has multiple input fields we probably want to test those too (doesn't seem much point checking one out of x inputs). This would at least be an improvement over what we currently do.

On the other hand we could go crazy and spend loads of effort implementing our own layer of security tests during the build phase (not a bad idea) but this would require a whole load of work.

I'd be interested to hear if you had any thoughts on this?

[finder-frontend]: https://github.com/alphagov/smokey/blob/9e0c307755e4590fb9af3a6270329b48b7d83817/features/finder_frontend.feature#L53-L59
[feedback]: https://github.com/alphagov/smokey/blob/a60df2cde5c886fa6d54e8cd820d90facca89e8e/features/feedback.feature#L11-L14

1pretz1:
Going to merge this PR as the follow up chat is unrelated to the actual change.

ChrisBAshton:
It looks to me like this shouldn't be a Smokey test - we should have comprehensive frontend tests in each frontend repo instead. Not sure how it ended up in here! I guess it's better to have it than not to have it, but ideally should just be a unit test in [feedback](https://github.com/alphagov/feedback/) IMO.

1pretz1:
Good point @ChrisBAshton. I suppose a smokey test may be a good idea if the malicious code is potentially moving through multiple applications (e.g. finder-frontend -> search-api) but acknowledge that the code should be escaped and handled by the frontend applications.

Does writing a card up to investigate standardising the way we test security across frontend apps seem a reasonable outcome out of this conversation? (happy for me to write this up)

ChrisBAshton:
That sounds good to me - thanks @1pretz1 !

PR #777, opened 2021-01-18
Add test environment

bilbof:
This allows us to run smokey against the test AWS environment. The following test passes:

```
AUTH_USERNAME=x AUTH_PASSWORD=y ENVIRONMENT=test bundle exec cucumber --profile test --strict-undefined -t @replatforming -t "not @notreplatforming"
```

This is required by the replatforming team, as we're trying to get smokey running in & against ECS.

### DIFF
--- a/config/cucumber.yml
+++ b/config/cucumber.yml
@@ -1,4 +1,5 @@
 default: -t "not @benchmarking" -t "not @pending" -t "not @local-network" -t "not @notintegration"
+test: -t "not @pending" -t "not @nottest" -t "not @notaws"
 integration: -t "not @pending" -t "not @notintegration" -t "not @notaws"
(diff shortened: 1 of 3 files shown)


config/cucumber.yml
```diff
@@ -1,4 +1,5 @@
 default: -t "not @benchmarking" -t "not @pending" -t "not @local-network" -t "not @notintegration"
+test: -t "not @pending" -t "not @nottest" -t "not @notaws"
```
richardTowers, line 2: Should this mention the `@replatforming` tag you added somewhere? Or do you want that to be explicit?

bilbof: Yeah I'd prefer it to be explicit. Partially for consistency, but also since we can use the `replatforming` tag against other environments, which might come in handy when testing against integration.

richardTowers (approved):

PR #778, opened 2021-01-19
Containerised smoke tests

bilbof:
This enables us to run the smoke tests from a container.

No build pipeline is set up for this yet. You'll need to manually build the container image with `docker build . -t smokey` then run the smoke tests like below.

Test environment (replatforming subset):

```
docker run -it -e AUTH_USERNAME=<user> -e AUTH_PASSWORD=<pass> -e ENVIRONMENT=test smokey bundle exec cucumber --profile test --strict-undefined -t 'not @benchmarking' -t @replatforming -t "not @notreplatforming"
```

Staging environment (smartanswers subset):

```
docker run -it -e AUTH_USERNAME=<user> -e AUTH_PASSWORD=<pass> -e ENVIRONMENT=staging_aws smokey bundle exec cucumber --profile staging_aws --strict-undefined -t 'not @benchmarking' -t @app-smartanswers
```

NB. Other environment variables are required (see [puppet for the list](https://github.com/alphagov/govuk-puppet/blob/84689ae2c36bbc7740da6a8d70c7f4e04d0e2357/modules/govuk/manifests/apps/smokey.pp)).

Changes to the test config:

- The additional timeout is required as I found the browsermob proxy would occasionally take over 10 seconds to start.
- The flag `--no-sandbox` was required in order to run Chrome as the root user in Docker. This is a shame but not a great risk given we're just making requests to our own site.
- Added environment variable flag `ENABLE_BROWSERMOB_LOGS` to make it easier to debug the image without needing to rebuild it.

The smoke tests expect to be run within our infrastructure (e.g. some tests make calls to non-public endpoints, such as publishing-api). We should mark those tests for removal, so we can run a subset of tests for CI for this repo (from concourse).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,37 @@
+FROM ruby:2.7.2
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential \
+  libpq-dev libxml2-dev libxslt1-dev dumb-init default-jre
+
+ENV APP_HOME /smokey
+RUN mkdir $APP_HOME
+
+ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE 1
+
+# Install Google Chrome
+RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -
+RUN sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list'
+RUN apt-get update && apt-get install -y google-chrome-stable && apt-get clean
+
+WORKDIR $APP_HOME
+ADD Gemfile* $APP_HOME/
+ADD .ruby-version $APP_HOME/
+RUN bundle install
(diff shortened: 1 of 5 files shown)


richardTowers (approved):
Looks good to me. I remember doing [something similar for the vulnerable people service](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/master/concourse/Dockerfile).

The Dockerfile advice used to be to reduce the number of layers in your images by merging the `RUN` steps. I'm not sure if that's still best practice - the docs suggest it might not be? - https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers

bilbof:
Aha, I should have checked that. Thanks, I removed a couple of `RUN` steps.

fredericfran-gds (approved):
looks good. I had a look at how other people run headless Chrome when root user. Not much easy options.

bilbof:
Thanks for having a look @fredericfran-gds. Yes, it's not great - we're doing the same thing for the e2e tests.

PR #781, opened 2021-01-21
Add .dockerignore file

bilbof:
This will improve the docker build performance, by excluding some files and directories not required by the container image.


sengi (approved):
I'd consider keeping it in lexicographical order but that's probably just me being fussy 😁

PR #782, opened 2021-01-22
Build & Push Smokey container image from Concourse

bilbof:
This adds a CI pipeline that:

1. detects new commits to the main branch
2. determines a new semantic version (version stored in its S3 bucket + 1 minor version)
3. builds a new container image
4. tags the image with `latest` and the new version
5. pushes the image to DockerHub
6. adds a tag to the commit on the GitHub repo, prefixed with "v" (e.g. v1.2.0)
7. updates S3 bucket containing the [semver-resource](https://github.com/concourse/semver-resource)

I expect we can copy this procedure for other pipelines - PaaS made their [integration test pipeline config](https://github.com/alphagov/paas-release-ci/blob/f6bef1a0cc3da5290e402506cf2e4e477817233f/pipelines/integration-test.yml) generic with variables - we can reuse the same config for multiple pipelines that do the same thing.

Tagging the commit is useful since we can watch for new version tags and trigger other builds.

### DIFF
--- a/concourse/pipelines/ci.yml
+++ b/concourse/pipelines/ci.yml
@@ -1,0 +1,78 @@
+---
+definitions:
+
+resources:
+  - icon: github
+    name: smokey
+    type: git
+    source:
+      uri: git@github.com:alphagov/smokey.git
+      branch: main
+      private_key: |
+        ((govukci_private_key))
+
+  - name: smokey-image
+    type: registry-image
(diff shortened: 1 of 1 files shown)


bilbof:
Moving this to [alphagov/govuk-infrastructure](https://github.com/alphagov/govuk-infrastructure).

PR #799, opened 2021-03-15
Remove performance platform

theseanything:
The performance platform has now been retired and redirected to data.gov.uk.


theseanything:
Use #800 instead

PR #805, opened 2021-03-16
Add a test for the account-api healthcheck

barrucadu:
This is to ensure that the app is running and can talk to postgres.

Based on https://github.com/alphagov/smokey/blob/main/features/travel_advice_publisher.feature

[Test run](https://deploy.integration.publishing.service.gov.uk/job/Smokey/27878/console)


huwd (approved):
:oof: is smokey all cucumber?

PR #828, opened 2021-06-07
Add publishing smoke test

karlbaker02:
This PR adds a smoke test for Publisher which creates a new draft `Place` and ensures it is then visible on the draft stack, before then proceeding to delete the draft.

This test will not run on existing production and will not run on the new replatformed production when it's ready, as it has been given the tag `@notproduction` (where the Smokey run for each environment by Concourse will exclude tags matching `@not<environment>`).

Trello: https://trello.com/c/9QipS0X0

### DIFF
--- a/features/publisher.feature
+++ b/features/publisher.feature
@@ -7,3 +7,37 @@ Feature: (Mainstream) Publisher
     When I request "/healthcheck/ready"
     Then JSON is returned
     And I should see ""status":"ok""
+
+  @notproduction
+  Scenario: Can create a draft place
+    Given I have the smokey run identifier
+    When I try to login as a user
+    And I go to the "publisher" landing page
+    And I go to add an artefact
+    And I create a draft place
+    And I preview the draft place
+    Then I should see that the content has been published
+
+  @notproduction
+  Scenario: Can publish a draft place to live
+    Given I have the smokey run identifier
+    When I try to login as a user
+    And I go to the "publisher" landing page
+    And I search for the content in category "Drafts"
+    And I go to the content from the search results
+    And I request 2nd pair of eyes
+    And I skip the review
+    And I publish the draft
+    And I view the published content on live
+    Then I should see that the content has been published
+
+  @notproduction
+  Scenario: Can unpublish a live place
+    Given I have the smokey run identifier
+    When I try to login as a user
+    And I go to the "publisher" landing page
+    And I search for the content in category "Published"
+    And I go to the content from the search results
+    And I go to the "Unpublish" section
+    And I unpublish the edition
+    Then I should see that the content has been unpublished
--- a/features/step_definitions/publisher_steps.rb
+++ b/features/step_definitions/publisher_steps.rb
@@ -1,0 +1,117 @@
+module Status
+  OK = 200
+  NOT_FOUND = 404
+  GONE = 410
+end
+
+When "I go to add an artefact" do
+  click_on "Add artefact"
+  expect(page).to have_content("New artefact")
+end
+
+When "I create a draft place" do
+  fill_in "Title", with: publication_title
+  fill_in "Slug", with: parameterize(publication_title)
+  select "Place", from: "Format"
+  click_button "Save and go to item"
+end
+
+When "I preview the draft place" do
+  click_on "Preview"
+
+  wait_until do
+    status = get_status_code do
+      visit page.current_url
+    end
+
+    status != Status::NOT_FOUND
+  end
+end
+
+When "I should see that the content has been published" do
+  expect(page).to have_content(publication_title)
+  expect(page.current_path).to eq(publication_path)
+  expect(page).to_not have_content("GOV.UK Publisher")
+end
+
+When /I search for the content in category "([^"]*)"/ do |status|
+  select "Nobody", from: "Assignee"
+  fill_in "Keyword", with: publication_title
+  select "Place", from: "Format"
+  click_button "Filter publications"
+  click_link status
+end
+
+When "I go to the content from the search results" do
+  click_link publication_title
+end
+
+When "I request 2nd pair of eyes" do
+  click_on "2nd pair of eyes"
+  click_on "Send to 2nd pair of eyes"
+end
+
+When "I skip the review" do
+  click_on "Skip review"
+
+  within "#skip_review_form" do
+    click_on "Skip review"
+  end
+
+  wait_until { find('span[title="Status"]').text == "Ready" }
+end
+
+When "I publish the draft" do
+  click_on "Publish"
+  click_on "Send to publish"
+end
+
+When "I view the published content on live" do
+  click_on "View this on the GOV.UK website"
+
+  wait_until do
+    status = get_status_code do
+      visit_path publication_url
+    end
+
+    status != Status::NOT_FOUND
+  end
+end
+
+When /I go to the "([^"]*)" section/ do |section|
+  click_link section
+end
+
+When "I unpublish the edition" do
+  accept_alert do
+    click_button "Unpublish"
+  end
+end
+
+When "I should see that the content has been unpublished" do
+  expect(page).to have_content("Content unpublished")
+
+  wait_until do
+    status = get_status_code do
+      visit_path publication_url
+    end
+
+    status == Status::GONE
+  end
+end
+
+def parameterize(str)
+  str.downcase.gsub(/\s/,'-')
+end
+
+def publication_title
+  @publication_title ||= "Smokey test draft place #{$smokey_run_id}"
+end
+
+def publication_path
+  @publication_path ||= "/#{parameterize(publication_title)}"
+end
+
+def publication_url
+  @publication_url ||= "#{ENV['GOVUK_WEBSITE_ROOT']}#{publication_path}"
+end
--- a/features/step_definitions/smokey_steps.rb
+++ b/features/step_definitions/smokey_steps.rb
@@ -39,6 +39,10 @@ Given /^I am an authenticated API client$/ do
   @authenticated_as_client = true
 end
 
+Given /^I have the smokey run identifier$/ do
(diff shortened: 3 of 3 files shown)


features/step_definitions/publisher_steps.rb
```diff
@@ -0,0 +1,48 @@
...
+When "I create a draft place" do
+  fill_in "Title", with: "Smokey test draft place"
+  fill_in "Slug", with: "smokey-test-draft-place"
+  select "Place", from: "Format"
+  click_button "Save and go to item"
+end
+
+When "I preview the draft place" do
+  # Wait 5 seconds before previewing, to give time for Router to update it's routes
+  sleep(5)
```
bilbof, line 15: I wonder if instead we could use `wait_until` (see: features/support/delay.rb) which might make this more robust (and potentially quicker!).

features/step_definitions/publisher_steps.rb
```diff
@@ -0,0 +1,48 @@
+When "I go to add an artefact" do
+  click_on "Add artefact"
+  expect(page).to have_content("New artefact")
+end
+
+When "I create a draft place" do
+  fill_in "Title", with: "Smokey test draft place"
+  fill_in "Slug", with: "smokey-test-draft-place"
```
bilbof, line 8: If smokey unpublishes a page like `gov.uk/smokey-test-draft-place`, can smokey then create another artefact with the same slug? 

I think we have unique constraint on the `slug` which might require you to use a `SecureRandom.hex` suffix.

features/publisher.feature
```diff
@@ -7,3 +7,22 @@ Feature: (Mainstream) Publisher
     When I request "/healthcheck/ready"
     Then JSON is returned
     And I should see ""status":"ok""
+
+  @replatformingnotproduction @notproduction
```
bilbof, line 11: Mentioned in https://github.com/alphagov/govuk-infrastructure/pull/329 - I'm not sure we'll need `replatformingnotproduction` given we have the `notproduction` tag.

bilbof (commented):
Thanks, nice work. I've added a couple of questions inline.

It would be great if we could figure out how to test that publishing a draft works. At the moment we're testing the draft stack, and a full publish would reveal problems with the live stack: content-store, router-api, router.

features/step_definitions/publisher_steps.rb
karlbaker02: I considered this at the time but actually I think you're right. Interesting that [`wait_until` was actually removed in Capybara v2.0](https://www.varvet.com/blog/why-wait_until-was-removed-from-capybara) though!

features/step_definitions/publisher_steps.rb
karlbaker02: Yep, it certainly can. It's not unpublishing as the draft is never actually published - it's simply deleting the edition of the draft, which removes the route from the draft stack ready to re-use it next time around.

karlbaker02: We should only ever have one of these drafts in existence at any point in time, unless it's possible for us to run multiple jobs of Smokey at once which would then have the potential for a race condition, at which point some random string might be a better bet.

features/publisher.feature
karlbaker02: Noted.

karlbaker02:
> Thanks, nice work. I've added a couple of questions inline.
> 
> It would be great if we could figure out how to test that publishing a draft works. At the moment we're testing the draft stack, and a full publish would reveal problems with the live stack: content-store, router-api, router.

That makes sense - was that your original intention when you wrote the card? I may have interpreted the output slightly differently, but very happy to amend to a) publish the draft and b) then remove what has been published.

features/step_definitions/publisher_steps.rb
bilbof: Yes I think we would want to run this smoke test in parallel (e.g. after deploys of the apps under test), so it would be great if we could find a way to enable this.

bilbof:
Yeah, sorry the requirements were not clear. Would be good to publish the piece of content and also enable smokey to run in parallel.

karlbaker02:
> Thanks, nice work. I've added a couple of questions inline.
> 
> It would be great if we could figure out how to test that publishing a draft works. At the moment we're testing the draft stack, and a full publish would reveal problems with the live stack: content-store, router-api, router.

This has now been done

features/step_definitions/publisher_steps.rb
```diff
@@ -0,0 +1,113 @@
...
+end
+
+def publication_path
+  @publication_path ||= "/#{parameterize(publication_title)}"
+  @publication_path
+end
+
+def publication_url
+  @publication_url ||= "#{ENV['GOVUK_WEBSITE_ROOT']}#{publication_path}"
+  @publication_url
```
bilbof, line 104: You can delete this line as the line above will return the same thing.

features/step_definitions/publisher_steps.rb
```diff
@@ -0,0 +1,113 @@
...
+  select "Place", from: "Format"
+  click_button "Save and go to item"
+end
+
+When "I preview the draft place" do
+  click_on "Preview"
+
+  wait_until do
+    visit page.current_url
+    page.has_no_text?(:all, /not found/i)
```
bilbof, line 18: Can you use the status code of `200` instead? I think this'll be less likely to change than the text.

features/step_definitions/smokey_steps.rb
```diff
@@ -39,6 +39,10 @@
   @authenticated_as_client = true
 end
 
+Given /^I have the smokey run identifier$/ do
+  $smokey_run_id ||= SecureRandom.hex
```
bilbof, line 43: Interesting solution! I guess if smokey crashes after publishing a doc but before deleting it, we'll have residual docs in our integration / staging environments. This is probably fine since they get wiped daily, no need to go further 🤔

features/step_definitions/publisher_steps.rb
```diff
@@ -0,0 +1,113 @@
...
+  @publication_url ||= "#{ENV['GOVUK_WEBSITE_ROOT']}#{publication_path}"
+  @publication_url
+end
+
+def page_contains_no_error?(error)
+  visit page.current_url
+
+  if page.has_no_content?(error)
+    return true
+  end
```
bilbof, line 112: Here can you do something like `page.has_no_content?(error)` ?

features/step_definitions/publisher_steps.rb
```diff
@@ -0,0 +1,113 @@
...
+    click_button "Unpublish"
+  end
+end
+
+When "I should see that the content has been unpublished" do
+  expect(page).to have_content("Content unpublished")
+
+  wait_until do
+    visit_path publication_url
+    page.has_text?(:all, /gone/i)
```
bilbof, line 84: Same here, it might be nice to use the status code we use for gone pages.

bilbof (commented):
Great work. It's a bit tightly coupled to the text of the pages under test, but that's pretty common for our smoke tests. There's also a risk we leave docs published in environments under test if unpublishing fails for some reason.

It might be nice to ensure that smokey can't publish in production, even if the `not @notproduction` flag is omitted. It's quite easy for one to forget to use this flag and point smokey at prod.
Perhaps there's a signon permission we can give to smokey in environments under test, but not in prod? Of course, we'd want failures due to this reason to be clear in error messages.

karlbaker02:
> Great work. It's a bit tightly coupled to the text of the pages under test, but that's pretty common for our smoke tests. There's also a risk we leave docs published in environments under test if unpublishing fails for some reason.
> 
> It might be nice to ensure that smokey can't publish in production, even if the `not @notproduction` flag is omitted. It's quite easy for one to forget to use this flag and point smokey at prod.
> Perhaps there's a signon permission we can give to smokey in environments under test, but not in prod? Of course, we'd want failures due to this reason to be clear in error messages.

Thanks for this comment, I missed it when working on the other changes last week. That's a fair point about ensuring that Smokey can't publish in production, even it the `not @notproduction` flag is omitted. The way I see that working is by the absence of the `skip_review` permission, which we've added to Smokey in other environments so that it can review its own test publications and see them all the way through to the live stack. We can also add in additional checks on the page, for example to ensure that we're not seeing the _Production_ environment label.

bilbof (approved):
⭐  Nice work

karlbaker02:
> > Great work. It's a bit tightly coupled to the text of the pages under test, but that's pretty common for our smoke tests. There's also a risk we leave docs published in environments under test if unpublishing fails for some reason.
> > It might be nice to ensure that smokey can't publish in production, even if the `not @notproduction` flag is omitted. It's quite easy for one to forget to use this flag and point smokey at prod.
> > Perhaps there's a signon permission we can give to smokey in environments under test, but not in prod? Of course, we'd want failures due to this reason to be clear in error messages.
> 
> Thanks for this comment, I missed it when working on the other changes last week. That's a fair point about ensuring that Smokey can't publish in production, even it the `not @notproduction` flag is omitted. The way I see that working is by the absence of the `skip_review` permission, which we've added to Smokey in other environments so that it can review its own test publications and see them all the way through to the live stack. We can also add in additional checks on the page, for example to ensure that we're not seeing the _Production_ environment label.

Given the fact that we currently don't have the environment label available in our `test` account due to the fact that we're not running `deploy.rb` from _govuk-app-config_ and [setting the appropriate configuration](https://github.com/alphagov/govuk-app-deployment/blob/c52a9c767dff658a912aa62cea5adde6d28ea94d/recipes/govuk_admin_template.rb) for the [archived `govuk_admin_template` gem](https://github.com/alphagov/govuk_admin_template/tree/master#environment-indicators), I'm going to for now skip the environment check. If/when we add this config into ECS or we switch to use the admin layout from _govuk_publishing_components_ then it will be worth revisiting this.

PR #835, opened 2021-06-29
Change SmartAnswer example for Imminence test

edwardkerry:
The "check-landlord-immigration" smart answer is no longer suitable for testing Imminence, so this swaps it for one that is.

There is now only one SmartAnswer that uses Imminence, the BenefitCapCalculator.

This commit removes a Scenario for stepping through an Imminence-enabled smart answer because:
* the benefit cap calculator only uses a post-code in the final step
* it is possible for the steps to fail validation but to still return a 200,
making this test unreliable

Instead we can now rely on the existing "Check postcode lookup" Scenario which actually asserts that the Imminence call has succeeded by checking page text.

[trello](https://trello.com/c/4mogKZrw/2584-temporarily-take-down-check-if-someone-can-rent-your-residential-property-remove-the-start-page-button-replace-the-start-page-te)

### DIFF
--- a/features/smartanswers.feature
+++ b/features/smartanswers.feature
@@ -27,16 +27,6 @@ Feature: Smart Answers
     | /vat-payment-deadlines/y/2000-01-31         |
     | /vat-payment-deadlines/y/2000-01-31/cheque  |
 
-  Scenario: Check stepping through a smart answer that uses Imminence
-    Then I should be able to visit:
-    | Path                                              |
-    | /landlord-immigration-check                       |
-    | /landlord-immigration-check/y                     |
-    | /landlord-immigration-check/y/SW1A2AA             |
-    | /landlord-immigration-check/y/SW1A2AA/yes         |
-    | /landlord-immigration-check/y/SW1A2AA/yes/yes     |
-    | /landlord-immigration-check/y/SW1A2AA/yes/yes/yes |
-
   Scenario: Check stepping through a smart answer that uses the Worldwide API
     Then I should be able to visit:
     | Path                                                |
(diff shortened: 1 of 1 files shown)


features/smartanswers.feature
```diff
@@ -29,13 +29,12 @@ Feature: Smart Answers
...
-    | /landlord-immigration-check/y/SW1A2AA             |
-    | /landlord-immigration-check/y/SW1A2AA/yes         |
-    | /landlord-immigration-check/y/SW1A2AA/yes/yes     |
-    | /landlord-immigration-check/y/SW1A2AA/yes/yes/yes |
+    | Path                                                                             |
+    | /benefit-cap-calculator                                                          |
+    | /benefit-cap-calculator/y                                                        |
+    | /benefit-cap-calculator/y/yes/no/none/child_benefit/100.0/400.0/single/SW1A2AA   |
+    | /benefit-cap-calculator/y/yes/no/none/child_benefit/100.0/400.0/couple/SW1A2AA   |
+    | /benefit-cap-calculator/y/yes/no/none/widow_pension/400.0/100.0/single/SW1A2AA   |
```
kevindew, line 37: Is there a particular logic for needing all these URLs? i.e. do they test different imminence features? I'm sort of wondering why we need these 5 URLS would say just the third meet the test?

If there's some reason to it, it'd be good to have it in the commit message.

By the way there's some other tests too: https://github.com/alphagov/smokey/blob/2398362070b4798a77c6c4c2e89d6c39d953d426/features/smartanswers.feature#L107-L108

kevindew (approved):
Thanks @edwardkerry 🎉

PR #836, opened 2021-07-01
Feedback component tests for each frontend app

kevindew:
This adds regression tests for the feedback component that is available
across GOV.UK frontend apps. At the time of writing this
component is currently broken in a number of applications and we've
found out that there wasn't tests available to catch the problem. A fix
is currently in progress for the problem [1] and these are being added
so that we can learn sooner if the problem re-occurs.

The feedback component has become particularly complex as it requires a
high degree of integration. The items involved are: the frontend
application, the feedback application, the slimmer gem, static application
and it's integration with govuk_publishing_components. This seems to leave
Smokey as the only tool that has a sufficient perspective to test this
component is integrated correctly. Because of this complexity I felt it
was necessary to test each application, as any one of those can
break while the component is still operational in other apps (this
is the case currently where this component is broken for 3 apps).

There was already a test that confirmed a user could submit a request
for the email survey from the homepage (frontend app). I
decided against extending this and instead wrote an additional
complimentary test. I chose this because: I thought it was worth testing
both the email survey and the report a problem functionality in a
consistent way; It looked like it'd be tricky to understand failures if
asserted on a post request; and I didn't want to bombard feedback with
lots of duplicate requests. I instead made this test into one that
exercises the routes feedback can take.

In order to exercise all the apps, and have them appropriately tagged, I had
to take this rather verbose approach of setting up lots of tagged
example tables. I'm not an expert on Gherkin by any means, so it could
be there's a simpler way to do this.

I've commented out the tests for applications that are currently failing
so that we can merge this in before to fix goes out. I'll plan to remove
the comments from them as the fix is deployed.

[1]: https://github.com/alphagov/slimmer/pull/271

### DIFF
--- a/features/feedback.feature
+++ b/features/feedback.feature
@@ -28,5 +28,83 @@ Feature: Feedback
     When I visit "/"
     And I click to say the page is not useful
     And I submit the email survey signup form
-    Then I see a signup confirmation message
+    Then I see the feedback confirmation message
     And a request is sent to the feedback app
+
+  Scenario: Check feedback component behaviour
+    When I visit "<url>"
+    And I confirm it is rendered by "<application>"
+    And I click to report a problem with the page
+    Then I see the report a problem form
+    When I close the open feedback form
+    And I click to say the page is not useful
+    Then I see the email survey signup form
+    When I close the open feedback form
+    And I click to say the page is useful
+    Then I see the feedback confirmation message
+
+    # Failing as of 30 June 2021
+    #
+    # @app-collections
+    # Examples:
+    #   | url                       | application |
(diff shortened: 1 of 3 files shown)


andysellick (approved):
Looks good to me. I assume we already test that the form submissions work properly somewhere?

kevindew:
Thanks for the review 👍 

> Looks good to me. I assume we already test that the form submissions work properly somewhere?

Yup, there's [this one](https://github.com/alphagov/smokey/blob/465c0752175e745398b137b9c895fe3bcc8c0037/features/feedback.feature#L27-L32) that sends a real request to Feedback.

PR #874, opened 2021-11-24
Add smoke test for elections API

sihugh:
Should give us a clue if things break, as it's usually quite low traffic so might go unnoticed for a while.

Relevant: https://docs.publishing.service.gov.uk/apps/frontend/elections-api.html


hannako (approved):
👍

PR #886, opened 2022-02-14
Bypass cache consistently for pages under test

benthorner:
https://trello.com/c/D2Yt6TrG/251-audit-tests-in-the-smokey-repo

This fixes several instances where we're not bypassing caching, and
yet I can see caching headers when manually requesting these pages.
Bypassing the cache means we can be confident that any changes to the
underlying frontends is picked up in a timely manner.

Scenarios that still don't bypass caching:

- "gov_uk_redirect.feature" - these are testing the CDN and I don't
observe any caching headers in the responses

- backend app features - cache bypass is not relevant to backend apps

- "mirror.feature" - I wasn't able to test this as I don't have access
to the Production mirrors

- draft environment authentication - these 302 responses appear to omit
caching headers (possibly an obscure Rails default)

- "ab_testing.feature", "caching.feature" - obviously not!

- assets /__canary__ route - this special route is not cached anyway [1]

I found these pages by auditing every scenario in Smokey and noting
the ones that weren't bypassing caching. The historic lack of cache
bypass raises a general question: should we bypass by default? I think
this makes more sense than testing Varnish by default.

A question to pursue in later work.

[1]: https://github.com/alphagov/govuk-puppet/blob/32c1bbbb10067078c1406170666a135b4a10aaea/modules/router/templates/assets_origin.conf.erb#L78

## Testing

## Deployment

Steps to deploy a change:

### DIFF
--- a/features/assets.feature
+++ b/features/assets.feature
@@ -16,16 +16,19 @@ Feature: Assets
 
   Scenario: Check an asset can be served
     Given I am testing "assets"
+    And I force a varnish cache miss
     When I request "/media/513a0efbed915d425e000002/120613_Albania_Travel_Advice_WEB_Ed2_jpeg.jpg"
     And I should get a content length of "212880"
 
   Scenario: Check assets with a docx extension are served correctly
     Given I am testing "assets"
+    And I force a varnish cache miss
     When I request "/media/59f70d5640f0b66bbc806ed3/questionnaire-for-accommodation-providers-online-hotel-booking.docx"
     Then I should get a "Content-Type" header of "application/vnd.oasis.opendocument.text"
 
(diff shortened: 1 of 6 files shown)


benthorner:
All modified features pass when testing on Integration ✅.

https://deploy.integration.publishing.service.gov.uk/job/Smokey/37708/console

kevindew (approved):
Nice, these look good to me 👍

benthorner:
Thanks @kevindew !

Would you mind deploying this to Production? I know I can merge and it'll get there automatically overnight, but in case it does cause a problem there ideally I'd deploy it manually so 2ndline are aware.

PR #889, opened 2022-02-22
Enable tests for replatformed applications

bilbof:
These tests should pass now that the apps are running in Kubernetes (`replatforming` tag). I removed this tag from some apps which we're not going to be running in a passing state for the next milestone of our project.

I've also added in the `app-<appname>` tag to some apps which were missing it.


fredericfran-gds (approved):
looks good

PR #891, opened 2022-02-24
Ignore Manuals Frontend smoke tests in EKS

bilbof:
This application is failing its smoke tests because the JS for this application is not loading.

This is expected as we have not yet completed the work to serve the static assets for this service.

Ignoring the tests will get the smoke test suite running only tests we expect to pass.


bilbof:
Closing as it's acceptable for these tests to be failing while we wait for the assets work to complete.

PR #892, opened 2022-03-03
Fix government-frontend test by normalising paths

bilbof:
The test at government_frontend.feature:78 relies on trailing slashes ("/")
being removed. This currently is done by Fastly VCL at
https://github.com/alphagov/govuk-cdn-config/blob/413abc509daea829d2eb8e125de711819b4d7409/vcl_templates/www.vcl.erb#L265

However, in cases where the Fastly CDN isn't fronting the site, we want
this test to still pass, so normalising the path by removing trailing slashes
here will make this more robust.

Tested:

```
ENVIRONMENT=production bundle exec cucumber features/government_frontend.feature:78

ENVIRONMENT=integration AUTH_USERNAME=<insert_username> AUTH_PASSWORD=<insert_pass> GOVUK_WEBSITE_ROOT="https://www-origin.eks.integration.govuk.digital/"  bundle exec cucumber features/government_frontend.feature:78
```

## Testing

## Deployment

Steps to deploy a change:

### DIFF
--- a/features/step_definitions/smokey_steps.rb
+++ b/features/step_definitions/smokey_steps.rb
@@ -228,10 +228,10 @@ end
 Then /^I should be at a location path of "(.*)"$/ do |location_path|
   if @response
     uri = URI(@response['location'])
-    expect(uri.path).to eq(location_path)
+    expect(normalise_path(uri.path)).to eq(location_path)
   else
     uri = URI(page.current_url)
-    expect(uri.path).to eq(location_path)
+    expect(normalise_path(uri.path)).to eq(location_path)
   end
 end
 
@@ -331,4 +331,9 @@ def get_status_code
   $proxy.new_har
   yield
   $proxy.har.entries.first.response.status
+end
+
+def normalise_path(path_str)
+  return path_str if path_str == "/"
+  path_str.chomp('/') # strip trailing slash
 end


features/step_definitions/smokey_steps.rb
```diff
@@ -228,10 +228,10 @@ def should_see(text)
 Then /^I should be at a location path of "(.*)"$/ do |location_path|
   if @response
     uri = URI(@response['location'])
-    expect(uri.path).to eq(location_path)
+    expect(normalise_path(uri.path)).to eq(location_path)
```
sengi, line 231: I would just write `uri.path.chomp('/')` inline - it's easier to read and unambiguous as to the intent.

features/step_definitions/smokey_steps.rb
```diff
@@ -332,3 +332,9 @@ def get_status_code
   yield
   $proxy.har.entries.first.response.status
 end
+
+def normalise_path(path_str)
+  return path_str if path_str == "/"
+  return path_str[0..-2] if path_str[-1] == "/" # strip trailing slash
```
benthorner, line 338: This will allow for a false positive [over here](https://github.com/alphagov/smokey/blob/main/features/gov_uk.feature#L12).

One alternative would be to have a separate step that's more permissive - could be specific to this scenario or the other one.

features/step_definitions/smokey_steps.rb
bilbof: Thanks - I've used `chomp` below, though `"/".chomp` would result in "/" which isn't 100% what we want (unless we normalise both the expected and observed path) so I've kept in the helper method.

sengi: Ohh sorry, I didn't twig that you want to remove only one trailing `/`. I think that makes the behaviour of the method a bit unintuitive though: `foo/` -> `foo` but `foo//` -> `foo/` - that's at least a little bit odd.

sengi: Or in other words, the method kinda needs unit tests 🙈

sengi: Sorry ignore me, GitHub wasn't showing the updated code hence my confusion. It's all good.

features/step_definitions/smokey_steps.rb
bilbof: Thanks @benthorner, that is a great catch, I appreciate your keen 👀. Serves me right for not running the whole suite in my testing. Your suggestion should work.

bilbof:
Closing this PR as we'll see this fixed by spinning up a Fastly domain for our EKS cluster and this task is out of scope.

PR #893, opened 2022-03-10
Remove `webdrivers` gem dependency

ollietreend:
The `webdrivers` gem is no longer required because supported GOV.UK runtime environments now supply a ChromeDriver for use by Selenium – e.g. govuk-docker and Jenkins CI workers.

Smokey also runs as a containerised application in the new EKS hosting environment that's currently being built. For that, I've updated the Dockerfile in this repository to install Chromium and a compatible ChromeDriver from the default package repo. That way, we'll be guaranteed to have a compatible browser and webdriver installed, and it should be possible for developers to run it on an M1 MacBook because Chromium supports both AMD64 and ARM64.

## Testing

✅ This has been tested in integration.

Output of the test run can be seen here:
https://deploy.integration.publishing.service.gov.uk/job/Smokey/38547/

## Deployment

Steps to deploy a change:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,21 +8,17 @@ RUN mkdir $APP_HOME
 
 ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE 1
 
-# Install Google Chrome
-RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -
-RUN sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list'
-RUN apt-get update && apt-get install -y google-chrome-stable && apt-get clean
+# Install Chromium and ChromiumDriver
+RUN apt-get update -qq && apt-get install -y chromium chromium-driver && apt-get clean
 
 WORKDIR $APP_HOME
 ADD Gemfile* $APP_HOME/
 ADD .ruby-version $APP_HOME/
 RUN bundle install
-ADD Rakefile $APP_HOME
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -8,21 +8,18 @@ RUN mkdir $APP_HOME
...
-RUN sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list'
-RUN apt-get update && apt-get install -y google-chrome-stable && apt-get clean
+# Install Chromium and ChromiumDriver
+RUN apt-get update -qq && apt-get install -y chromium chromium-driver && apt-get clean
 
 WORKDIR $APP_HOME
 ADD Gemfile* $APP_HOME/
 ADD .ruby-version $APP_HOME/
 RUN bundle install
 ADD Rakefile $APP_HOME
```
kevindew, line 18: We'll need to delete this line (docker build crashes for me)

kevindew (commented):
This looks good to me aside from the comment I've flagged. Shall we check with #replatforming as to whether they want to run a test of it in their EKS environment?

I've given it a run locally (after making a quick amend) and it seems to be working fine.

I did this with:

```
docker build . -t smokey
 docker run -it -e AUTH_USERNAME=<username> -e AUTH_PASSWORD=<password> -e ENVIRONMENT=integration smokey bundle exec cucumber --profile test --strict-undefined -t 'not @BenchMarking'
```

and I'm only seeing tests fail that either require internal network access or signon credentials

Dockerfile
ollietreend: Nice catch – thanks

richardTowers (approved):

PR #908, opened 2022-03-29
Remove unnecessary @benchmarking tag

benthorner:

This stops the scenarios in benchmarking.feature from running when
we do a full run of Smokey. However, the tests are still run by the
Smokey Loop. Since we're removing the single alert associated with
a full run of Smokey [^1], the impact of these tests failing is low
and we may as well just run them.

Long term we should reconsider the need for these specific tests:
it's unclear why most of them exist.

In practice I don't think the failures will cause a problem for full runs 
of Smokey in Integration or Staging, compared to other failures:

Screenshot 2022-03-29 at 17 33 39

[^1]: https://github.com/alphagov/govuk-puppet/pull/11610

## Testing

## Deployment

Steps to deploy a change:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,4 +31,4 @@ ENTRYPOINT ["/usr/bin/dumb-init", "--"]
 # Users of the image are expected to override the default CMD and set env vars
 # ENVIRONMENT, AUTH_USERNAME, AUTH_PASSWORD, and a profile flag.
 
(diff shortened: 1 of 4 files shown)


benthorner:
Tested and works on Integration (all benchmarking scenarios pass) ✅ .

ChrisBAshton (approved):

PR #910, opened 2022-04-04
Remove tests for HMRC 'service sign in' pages

ollietreend:
## Changes made

Following the deployment of alphagov/publisher#1573, these tests have become redundant. I've also cleaned up the associated step definitions which are now unused.
    
There are no longer any live services that offer the choice of signing in via either Government Gateway or GOV.UK Verify. So rather than adapt these tests to use a different example service, it made sense simply to remove them.

Also, my editor auto-fixed whitespace indentation in the changed `.feature` file. But I've included that as a separate commit to make the diffs easier to read.

## Testing

## Deployment

Steps to deploy a change:


kevindew (approved):
Nice one

benthorner:
Great to see these cleaned up 💯.

I had 3 out of 4 of them marked for removal, irrespective of the feature existing (I'm doing an audit of Smokey).

PR #915, opened 2022-04-08
Remove unused step definitions

benthorner:
Based on the output of:

    ENVIRONMENT=staging bundle exec cucumber --dry-run -f stepdefs -P

Tested and works with a full run of Smokey in Integration ✅.

Some scenarios (and thus steps) don't run in Integration, but I trust that the above command correctly identified steps that are genuinely unused, noting the use of the `-P` flag to disable all scenario filters i.e. all scenarios were run.


kevindew (approved):
Nice work

PR #916, opened 2022-04-08
Remove flakey tests for publisher app

benthorner:
Originally added in [^1].*

These tests have several major issues and add little value:

- They are significantly more flakey than all other Smokey tests,
generating between 20% and 40% of all new failures each day.

- Unlike all other tests, they can only run in Integration as they
make visible changes to GOV.UK. This creates a confusing precedent.

- They were added in response to an incident action [^2][^3], but
the action calls for tests in Production, where these cannot run.

- Unlike all other tests, they are order-dependent: the draft must
be created before it's published and then unpublished.

We should have end-to-end tests of the publishing side of GOV.UK,
but we need those tests to be reliable, which these are not. This
is a hard problem to solve and merits more time than we had when
we put these in, so we can solve it well - ideally in Production.

It's unlikely we'll be able to make these tests more reliable in
the near future. Leaving them in also poses a risk as they set a
number of precedents that we don't want other tests to copy, such
as order dependence and only running in Integration.

On balance, these tests create more issues than the value they add.
This commit removes them and their associated steps.

*Note that the card these were added against is very different to
the one directly linked to the incident action [^2]. It looks like
there was some misunderstanding at the planning stage of this work.

## Analysis of failures

Screenshot 2022-04-08 at 10 59 22

_Created by customising one of the panels in [this dashboard](https://grafana.integration.publishing.service.gov.uk/dashboard/file/icinga_alerts.json). The data we have is based on Icinga service events, which means we can’t tell how long the failures persisted for - only when new failures occur._

[^1]: https://github.com/alphagov/smokey/pull/828
[^2]: https://trello.com/c/XbzUB6l9/179-prober-and-alert-for-basic-end-to-end-publishing-journey
[^3]: https://docs.google.com/document/d/12tTb4o1Vn8qF-htppPP2qmei8UHQL9cDZ0DG0YWRZFA/edit#heading=h.p3b9bm9twi19


kevindew (approved):
Seems good to me, they're quite tangential to the rest of the suite. I agree with your assessment in the description.

I'd be pleased to get a review from someone from replatforming too.

sengi (approved):
SGTM

ChrisBAshton (approved):
The reasoning makes sense to me 👍

PR #918, opened 2022-04-12
Try removing rate limit bypass token everywhere

benthorner:
https://trello.com/c/DL5omRGP/230-smokey-needs-java-and-a-glitchy-proxy-server-to-run-tests

This is potentially unnecessary given many of the tests now probe
via the CDN, and the low frequency these run at in general means
it's highly unlikely it will trip any reasonable rate limiting at
our origin servers or our publishing backend servers.

Tested and works in Integration (Jenkins Job) ✅
Tested and works in Integration (Smokey Loop) ✅ 

There's still a risk that this will cause an issue in another environment 
over a sustained period, but I think it's worth the risk to merge and 
revert if we find this happens in practice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,6 @@ You can use the following environment variables to configure the tests:
 * `ENVIRONMENT`: used to set environment variables for [Plek](https://github.com/alphagov/plek)
 * `SIGNON_EMAIL`: email address of a user with a Signon account in the environment the tests are being run in
 * `SIGNON_PASSWORD`: password of a user with a Signon account in the environment the tests are being run in
-* `RATE_LIMIT_TOKEN`: a token used to bypass the default rate limiting
 
 **Note: you will need to be connected to the VPN to test against Integration or Staging.**
 
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -42,11 +42,6 @@ proxy = server.create_proxy
 $proxy = proxy
 
 # Add request headers
-if ENV["RATE_LIMIT_TOKEN"]
-  proxy.header({ "Rate-Limit-Token" => ENV["RATE_LIMIT_TOKEN"] })
-end
-
-
 if ENV["ACCOUNT_AUTH_USERNAME"] && ENV["ACCOUNT_AUTH_PASSWORD"]
   proxy.basic_authentication(
     "www.account.staging.publishing.service.gov.uk",
--- a/features/support/http_requests.rb
+++ b/features/support/http_requests.rb
@@ -90,10 +90,6 @@ def do_http_request(url, method = :get, options = {}, &block)
   if options[:cookies]
     headers["Cookie"] = options[:cookies].map { |k, v| "#{k}=#{v}" }.join("; ")
   end
-  rate_limit_token = ENV['RATE_LIMIT_TOKEN']
-  if rate_limit_token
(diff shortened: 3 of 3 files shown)


kevindew:
I agree about seeing whether this causes problems or not, do you plan to let 2nd line and/or platform reliability know?

ChrisBAshton (approved):
Ben's message from Slack:

> What I'll do for now is deploy it to Smokey Loop on Integration for a day and see if the longer run turns up anything obvious in logs or the dodgy Icinga metrics we do have - I'd expect to see some kind of new pattern if there's a problem.
If I don't find anything, then I think we could reasonably merge it in and let 2ndline know.

Sounds good to me 👍 approving on that basis.

benthorner:
## Short version

Further testing confirms this change will not cause a degradation in our monitoring.

I will still drop a note to 2ndline to let them know this will apply from tomorrow (after merging).

## Long version

Deploying this on Integration for a day didn't lead to any significant change in failures:

| Icinga (new failures) | Kibana (each failure) |
| - | - |
| Screenshot 2022-04-21 at 09 31 19 | Screenshot 2022-04-21 at 09 20 49 |

Kibana is more useful as we get a new log every time a feature fails. By elimination I confirmed the set of features that failed in the past day (getting to "no results") and then looked at patterns for each of them over the last week.

| | |
| - | - |
| Screenshot 2022-04-21 at 09 26 23 | Screenshot 2022-04-21 at 09 26 44 |
| Screenshot 2022-04-21 at 09 26 57 | Screenshot 2022-04-21 at 09 27 11 | 
| Screenshot 2022-04-21 at 09 27 27 | Screenshot 2022-04-21 at 09 27 47 |
| Screenshot 2022-04-21 at 09 28 08 | Screenshot 2022-04-21 at 09 28 51 | 

_(It looks like we stop getting logs for Smokey Loop between around 11PM and 9AM each day. I haven't tracked down what's causing this but it's consistent so not a cause for concern here.)_

Again, none of these show any significant change without rate limit bypass. It's still possible the removal caused a problem though: the failures for feedback and government_frontend are constant, so the reason for the failures could change without affecting the graphs. Looking at these two in detail shows the point of failure is at least the same:

| feedback (before trial) | feedback (during trial) |
| - | - |
| Screenshot 2022-04-21 at 09 42 12 | Screenshot 2022-04-21 at 09 41 42 |

_(there's something wrong with the logging for government_frontend: 1 failure is reported but none of the steps are showing as `[FAIL]`; it's at least consistent before and during the trial)_

This is _still_ not conclusive that the removal is benign, but it does show that it makes no difference to the quality of the alerts we're currently getting i.e. bypass has no benefit given the general flakiness of Smokey.

One final note: the only feature that wasn't tested here is [`mirror.feature`, which only runs in Production](https://github.com/alphagov/smokey/blob/main/features/mirror.feature#L5). This change won't affect that feature because the rate limit bypass only applies to Nginx origins.

benthorner:
## Update

If we ever need to reinstate this approach, then I found a way to do it without the need for a proxy. Selenium v4 gives us direct access to Chrome devtools, which means we can do something like this:

```
driver = Selenium::WebDriver.for :headless_chrome

driver.browser.devtools.execute_cmd(
  'Network.setExtraHTTPHeaders',
  headers: {'Rate-Limit-Token': 'xyz-secret'}
)

PR #924, opened 2022-04-28
Replace BrowserUp proxy with Chrome browser logging

benthorner:
Pays down: https://trello.com/c/DL5omRGP/230-smokey-needs-java-and-a-glitchy-proxy-server-to-run-tests

## Background to this PR

This is the final PR in a sequence to remove this thing:

- BrowserUp (formerly BrowserMob) was originally added to support
basic auth, which is no longer necessary [^1][^2].

- It got reused for rate limiting [^3] and JS error management [^4],
both of which have since become redundant.*

*If we do need to manage JS errors in future, we should be able to
do this using the Capybara Log Collector config [^5].

Having to manage a separate Java process is inherently error-prone,
as the (now redundant) troubleshooting guide explains.

## Spying on AJAX requests

It turns out headless Chrome can be configured to expose the same 
kind of data you can see manually in the dev inspector, which is 
sufficient to replace the proxy as a means of spying on requests.

Please review commit by commit.

## Testing

[Tested and works on Integration Jenkins](https://deploy.integration.publishing.service.gov.uk/job/Smokey/39731/console) ✅.

[^1]: https://github.com/alphagov/smokey/pull/913
[^2]: https://github.com/alphagov/smokey/pull/923
[^3]: https://github.com/alphagov/smokey/pull/918
[^4]: https://github.com/alphagov/smokey/pull/921
[^5]: https://github.com/dbalatero/capybara-chromedriver-logger#configuration


KludgeKML (approved):
Looks good - works well on local dev machine too, which previous version absolutely does not.

PR #931, opened 2022-05-09
Add Sentry / StatsD reporting for scenario errors

benthorner:

Supersedes: https://github.com/alphagov/smokey/pull/911
Depends on: https://github.com/alphagov/govuk-puppet/pull/11661, https://github.com/alphagov/govuk-secrets/pull/1284

Tested and works with errors in individual steps, as well as with
the "After" hook to detect JS errors. Backtraces for both types of
error usually include the step and its Ruby definition; we also tag
the feature and the scenario to help identification and filtering.
Sentry does a good job of expanding backtraces to show all the code.

Note: this is intended to be a trial of using Sentry to help identify common 
issues with Smokey. The intention is to run it for a week or so and then 
iterate or remove what's here based on its usefulness.

Tested (together with [Puppet](https://github.com/alphagov/govuk-puppet/pull/11661)) and works on Integration ✅.

## Screenshots

| Failed assertion | JS error on page |
| - | - |
| screencapture-sentry-io-organizations-govuk-issues-3259185495-events-7e14d55e5aa34ad69687d0a9471d9753-2022-05-09-11_39_53 | screencapture-sentry-io-organizations-govuk-issues-3259185495-2022-05-09-11_37_39 |

## Deployment

Steps to deploy a change:


features/support/hooks.rb
```diff
@@ -5,22 +5,16 @@
...
   # unformatted JavaScript errors from making their way into the Smokey loop's
   # JSON output file and breaking it.
   collector = Capybara::Chromedriver::Logger::Collector.new(log_destination: $stderr)
   collector.flush_and_check_errors!
+rescue Capybara::Chromedriver::Logger::JsError => e
+  if not $fail_on_js_error
+    puts "Detected JS error, but ignored it. #{e}"
+  else
+    raise
+  end
```
ChrisBAshton, line 25: Nice 👍

features/support/hooks.rb
```diff
@@ -4,7 +4,12 @@
   $fail_on_js_error = true
 end
 
-After do
```
ChrisBAshton, line 7: Hmmm the two `After` blocks are a little confusing. I think a quick comment above each of them would aid the reader, e.g.:

```ruby
# Check for backend errors
After do |scenario|
  # ...

# Check for frontend errors
After do |scenario|
  # ...
```

ChrisBAshton (approved):
LGTM ⭐

features/support/hooks.rb
benthorner: Sure, I've done this in https://github.com/alphagov/smokey/pull/931/commits/8f56cee879ac8d0bb1389bfb656cc794e87a25f7.

benthorner:
@ChrisBAshton I had to do some fairly major rebase work on the first commit.

Would you like to take another look?

features/support/env.rb
```diff
@@ -21,6 +21,9 @@
 # Set up basic URLs
 Capybara.app_host = ENV["GOVUK_WEBSITE_ROOT"]
 
+# Set up optional error reporting (using SENTRY_CURRENT_ENV for the environment)
```
ChrisBAshton, line 24: I'd drop the word 'optional' as it implies it can be configured per-test, whereas really we're opting the entire test suite into Sentry error reporting. (It may be optional whether we do it or not, but as we've decided to do it, I don't think there's value in this word 😁 )

```suggestion
# Set up error reporting (using SENTRY_CURRENT_ENV for the environment)
```

ChrisBAshton (approved):
LGTM ⭐ one nitpick comment but that's it.

features/support/env.rb
benthorner: Ah yes I see what you mean.

Removed in https://github.com/alphagov/smokey/pull/931/commits/2c950649bb8a769beb917eee2b360ba66efbb99e.

PR #933, opened 2022-05-09
Inline features from capybara chromedriver logger

KludgeKML:
## Testing

## Deployment

Steps to deploy a change:


features/support/log_message.rb
```diff
@@ -0,0 +1,62 @@
+class LogMessage
```
benthorner, line 1:
```suggestion
class ChromeBrowserLog
```

This is more precise in the general context of the Smokey repo.

features/support/hooks.rb
```diff
@@ -8,17 +8,45 @@
...
+  end
+
+  errors
+end
+
+def should_filter?(message)
+  %i[debug info warning].include?(message.level.downcase.to_sym)
+end
+
+def logs(type)
```
benthorner, line 45:
```suggestion
def browser_logs(type)
```

This is more precise in the general context of the repo.

benthorner (commented):
Works locally for me ✅.

Worth noting this is a prerequisite to avoid deprecation warnings in https://github.com/alphagov/smokey/pull/894

## Testing

Testing this is a bit tricky. I had to concoct a feature and point it at a local app, where I injected a JS error with a nice stacktrace: `this.foo()`.

Note that using `page.execute_script` doesn't work the same way: the JS exception is detected inline and not by the code we're changing here.

## Behaviour

This change does a good job of keeping the existing functionality. However, we do end up importing a fair - but still reasonable - chunk of code.

I think we can go even more minimal e.g.

```
  errors = logs(:browser)
    .select { |log| log.level == 'SEVERE' }
    .map(&:message)

  errors.each(&$stderr.puts)
  return unless errors.any?

  raise LogMessage::JsError,
    "Got some JS errors during testing:\n\n#{errors.join("\n")}"
```

Trying this locally shows printing the raw message is only marginally different to all the formatting that's done by the gem (that we're importing here).

Before:

```
      Got some JS errors during testing:
      
      ...
      📜   severe  http://host.docker.internal:6012/static/javascripts/all.js?251b630fa82bca9b64ae1c6634a0417d 13:1185
           Uncaught TypeError: this.foo is not a function
      📜   severe 
           http://host.docker.internal:6012/fonts/bold-tabular-b89238d840-v1.woff2 - Failed to load resource: the server responded with a status of 404 (NOT FOUND)
      ...
      ./features/support/hooks.rb:26:in `flush_and_check_errors'
      ./features/support/hooks.rb:7:in `After'
```

After:

```
      Got some JS errors during testing:
      
      ...
      http://host.docker.internal:6012/static/javascripts/all.js?251b630fa82bca9b64ae1c6634a0417d 13:1185 Uncaught TypeError: this.foo is not a function
      http://host.docker.internal:6012/fonts/bold-tabular-b89238d840-v1.woff2 - Failed to load resource: the server responded with a status of 404 (NOT FOUND)
      ...
      ./features/support/hooks.rb:26:in `flush_and_check_errors'
      ./features/support/hooks.rb:7:in `After'
```

Apart from the pointless icon, the main difference is that the message is all one line, whereas before the error was on a separate line. I don't think we get these often enough to need to optimise how they appear, so long as the error is in there somewhere.

What do you think?

## Stretch goal

You could use the new method to replace [this line](https://github.com/alphagov/smokey/blob/2017266b906ec7f8f96fb164cf3d8ba03c587296/features/step_definitions/datagovuk_steps.rb#L33), which would help avoid [this kind of problem](https://github.com/alphagov/smokey/blob/2017266b906ec7f8f96fb164cf3d8ba03c587296/features/support/browser.rb#L40-L42). Note: [this handy picture](https://stackoverflow.com/questions/35014990/because-it-violates-the-following-content-security-policy-directive-style-src) confims that CSP log is indeed "SEVERE".

KludgeKML:
I think I've covered all the required changes, that simplified version does seem to work and lets us remove the entire ChromeBrowserLog file/class (except for the vestigial JsError, which we can just inline in hooks.rb).

For the stretch goal, if I'm thinking about this right does it just mean we can ignore the entire step? If a CSP violation is severe, then the after-suite will throw up an error after that scenario runs even without the check step. We could mention that in the scenario description or something just to be explicit that what we're looking for is an SCP violation, but it's reasonable to fail that test if any other JS Severe turns up anyway, right?

It's not clear whether I'm testing that locally, because it looks like it's marked with @notintegration @notstaging. I guess I have to point to production to try that test?

benthorner (approved):
Tested and confirmed this works locally ✅.

Good point about the stretch goal. You could go either way with that:

- Change the feature to be focussed on previewing organograms (with JS error detection being incidental - probably don't need to mention it except in the commit).
- Just replace the line with the method we now have in hooks.rb (being careful with the memoisation).

Or just leave it.

This was always a stretch goal so I'll let you decide.

KludgeKML:
I think in this case what I might do is leave as is, but open an issue for myself to come back to it later (I really need to start unblocking the deployment of rate-limiting this week, so I'll make this a sort of vague neutral tech debt).

PR #936, opened 2022-05-10
Add guidance for installing Ruby / Chromedriver

benthorner:
This repo isn't covered by GOV.UK Docker, which takes care of both
of these setup concerns. My quick attempt at adding it was partly
successful, but when running the tests the latter ~half fail e.g.

    Then I should be able to visit:                 # features/step_definitions/smokey_steps.rb:162
      | Path                           |
      | /government/announcements.atom |
      | /government/statistics.atom  |
      invalid session id (Selenium::WebDriver::Error::InvalidSessionIdError)

For now it makes sense to at least offer guidance on running this repo
natively. Hopefully we can add it to GOV.UK Docker in future though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,30 @@ The tests also run in [a continuous Smokey loop](https://github.com/alphagov/gov
 
 ## Installation
 
+### chromedriver
+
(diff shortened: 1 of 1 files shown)


KludgeKML (approved):
LGTM

PR #938, opened 2022-05-10
Use GOV.UK Docker for local development

benthorner:
Relates to: https://github.com/alphagov/govuk-docker/pull/597

This simplifies the setup required, with only a minor adjustment
needed to the command to run [^1]. In future we could also suggest
using dotenv to avoid putting credentials in commands. I've also
tweaked the headings so this repo consistent with other apps [^2].

To test locally:

- Pull the GOV.UK Docker changes
- Run "make smokey"
- Run the tests as described

[^1]: https://github.com/alphagov/govuk-docker/blob/main/docs/how-tos.md#how-to-set-environment-variables
[^2]: https://github.com/alphagov/content-publisher#technical-documentation


ChrisBAshton (approved):
Tested and works locally 👍 I did have some test failures because of Signon 2FA screens (despite applying my 2FA code in an earlier browser session) but that shouldn't stop the merging of this. 1161 steps ran successfully for me, 42 were skipped, and 24 failed.

PR #939, opened 2022-05-17
Change puts to log to fix cucumber deprecation

KludgeKML:
## Testing


benthorner (approved):
Nice one 💯

PR #940, opened 2022-05-17
Add rate limit token support through chrome devtools

KludgeKML:
## Testing

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem "plek", "~> 4"
 gem "rest-client", "~> 2"
 gem "rspec", "~> 3"
 gem "selenium-webdriver", "~> 4"
+gem "selenium-devtools"
 
 group :development do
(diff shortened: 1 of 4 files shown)


benthorner (approved):
Thanks for all the effort getting to this point 🏅 💯.

I've tested this locally and confirmed it works as expected using features/feedback.feature:21 and copy / pasting the steps a few times i.e.

```
  Scenario: Check "is this page useful?" email survey
    When I visit "/"
    And I click to say the page is not useful
    And I submit the email survey signup form
    Then I see the feedback confirmation message
    And a request is sent to the feedback app
    When I visit "/"
    And I click to say the page is not useful
    And I submit the email survey signup form
    Then I see the feedback confirmation message
    And a request is sent to the feedback app
    ...
```

Running without this change, I consistently get an error (due to rate limiting). With this change, the modified scenario consistently passes.

PR #945, opened 2022-05-18
Include scenario in Sentry issue fingerprint

benthorner:

The Sentry plan we use doesn't provide any export / reporting tools,
which makes it hard to analyse failures of unrelated scenarios in the
same feature. Including the scenario makes analysis practical within
Sentry itself, assuming most scenarios fail for the same reason each
time. Note that this may lead to lots of issues if entire apps fail,
but that arguably reflects their flakiness being a priority to fix.

## Testing


benthorner:
Closing as I found a way to do reporting in Sentry 🎉.

PR #948, opened 2022-05-19
Fix duplicate Sentry issues for request errors

benthorner:

When a request to visit a page fails we get:

- A normal scenario failure, which was picked up by the shorter
"After" block. This reports an issue to Sentry as expected.

- A "SEVERE" browser log for the failed request, which was picked
up by the long, async "After" block. This leads to another issue.

The "async" block is the one that needs fixing: it should only be
picking up on "out-of-band" JS errors; that is, when we visit a
page and see some red in the console, despite the general content
on the page being sufficient for the scenario to pass.

Having two "After" blocks was making it hard to explain the control
flow, so as well as fixing the problem I've also combined them into
one as they both perform similar error reporting tasks.

Testing
=======

I didn't spot this in my previous tests as I was using artificial
scenarios. It turns out these weren't representative of the common
case, so for this commit I tested with something like this:

    When "whatever" do
      visit "local server with JS errors"
      raise "pretend failure"
    end

My test confirms the fix: before the commit I see two Sentry issues;
after the commit I only see one.

### DIFF
--- a/features/support/hooks.rb
+++ b/features/support/hooks.rb
@@ -2,8 +2,14 @@ Before do
   $fail_on_js_error = true
 end
 
-# Check for async JS failures
 After do |scenario|
+  # Scenario failed, so report the failure to Sentry
+  if scenario.exception
+    capture_error(scenario, scenario.exception)
+    next
+  end
+
+  # Scenario passed, but could have async JS errors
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
Makes sense 👍

PR #952, opened 2022-05-19
Fix cors problem reinstate rate limit support

KludgeKML:
Reinstates the rate limit header support, with web security turned off to avoid CORS errors.

When you make a cross origin request, there are certain headers you’re
allowed to send with it, in an allowlist. If you send any other header,
the system assumes you’re trying to authenticate yourself somehow or do
other dangerous cross-origin things. Our our rate-limit-header is treated
like that, so we'll get CORS errors. They're not important, so we just
turn off web security to remove them (before we filtered them out, as:
https://github.com/alphagov/smokey/pull/921)

Details about what --disable-web-security does: 
https://groups.google.com/a/chromium.org/g/chromium-discuss/c/Wms-ZaAwA1Q
https://github.com/chromium/chromium/blob/main/content/public/common/content_switches.cc#L280

Has been tested on:
- Integration (https://deploy.integration.publishing.service.gov.uk/job/Smokey/40071/console) (which includes rate-limited feedback)
- Staging (https://deploy.blue.staging.govuk.digital/job/Smokey/19165/console)
- Production (https://deploy.blue.production.govuk.digital/job/Smokey/11804/console)

Note that this branch requires a relatively recent chromedriver, but monitoring and jenkins machines in all environments have been updated to Chromedriver 101

## Testing

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem "plek", "~> 4"
 gem "rest-client", "~> 2"
 gem "rspec", "~> 3"
 gem "selenium-webdriver", "~> 4"
+gem "selenium-devtools"
 
 group :development do
   gem "pry-byebug", "~> 3"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -163,6 +163,8 @@ GEM
       rspec-support (~> 3.11.0)
     rspec-support (3.11.0)
     rubyzip (2.3.2)
+    selenium-devtools (0.101.0)
+      websocket (~> 1.0)
     selenium-webdriver (4.1.0)
       childprocess (>= 0.5, < 5.0)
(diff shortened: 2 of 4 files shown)


benthorner (approved):
So pleased this wasn't a deal breaker 🎉.

I've done a quick test of this locally just to be sure it (a) still fixes the rate limiting issue ✅ and (b) doesn't lead to failures due to CORS (tested with finder_frontend.feature) ✅.

Worth noting that `--disable-web-security` isn't documented particularly well, but what references I can find to it all agree it disables CORS checks [^1][^2].

[^1]: https://groups.google.com/a/chromium.org/g/chromium-discuss/c/Wms-ZaAwA1Q
[^2]: https://github.com/chromium/chromium/blob/main/content/public/common/content_switches.cc#L280

PR #964, opened 2022-06-01
Remove Manuals Frontend tests

1pretz1:
We've migrated manuals to Government Frontend and so these tests need
to be updated.

Passing Smokey tests using this branch: https://deploy.integration.publishing.service.gov.uk/job/Smokey/40290/ 

## Testing


hannako (approved):
👍

PR #965, opened 2022-06-01
Add guidance on what tests belong in this repo

benthorner:

**Deadline for discussion: 2022-05-15**

## Overview

In this PR we're continuing [the work we started in govuk-rfcs#128](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-128-continuous-deployment.md#safety-checks),
which established baselines for all the other ways we test GOV.UK.
We committed to do the same for Smokey [^1].

In this PR I'm trying to solve two hard problems:

1. Say when a test belongs in this repo, and when it doesn't.
2. Apply (1) to all the existing tests to show it makes sense.

To show (2), I've annotated all the tests that will be removed if
we merge this PR - immediately or after we improve in-app tests.

We need to be ruthless when we add smoke tests:

- Smokey is used for alerts, so each test should represent something
that is truly critical to GOV.UK as a whole.

- Smokey tests are flakey because they run against real environments.
The more we have, the more flakiness there will be.

## Consequences

There are currently 154 scenarios in Smokey. Some of these are
"Scenario Outline"s and actually expand to even more tests. I'll
just refer to both of them as "tests" for simplicity.

Consequences:

1. Fewer tests. We will immediately remove 69 tests. We will also
remove 9 tests once sufficient tests exist elsewhere.

2. More reliable. Removing tests will remove or reduce the impact
of flakey tests on deployments and alerts [^2].

3. Faster tests. Deleting all the REMOVE and EXPORT tests takes 
the total run time down from ~4 mins to ~2 mins.

4. GOV.UK strategy. This guidance is the missing piece we need to
have [a single document that says "How we test GOV.UK"](https://github.com/alphagov/govuk-developer-docs/pull/3564).

## How to review

1. 👉 [Review the guidance (rendered version)](https://github.com/alphagov/smokey/blob/writing-tests/docs/writing-tests.md)*.
2. 👉 Check apps you work on for actions in the diff.
3. ✅ Approve or 👍 this PR if you agree with 1 + 2.

*Note: we don't currently have any "Tests that make changes to GOV.UK" in Smokey.

Be prepared to feel uncomfortable:

- Removing tests can feel wrong. I've tried to help you by adding
explanations and links to supporting evidence.

- Shout if we're making a mistake. If you have evidence that an
action will put GOV.UK at risk, leave a comment.

[^1]: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-128-continuous-deployment.md#explainable-smoke-tests
[^2]: https://trello.com/c/S1e8GWWG/234-smokey-tests-are-highly-flakey

### DIFF
--- a/features/assets.feature
+++ b/features/assets.feature
@@ -3,6 +3,20 @@ Feature: Assets
   Tests for assets and asset-manager, including draft assets which are behind
   signon.
 
+  # TODO: REMOVE this test as it does not meet the elgibility
+  # criteria in docs/writing-tests.md.
+  #
+  # - Covers site-wide config: N (already covered)
+  # - Targets data transfer: N (not applicable)
+  # - Second critical check: N (not applicable)
+  #
+  # This test checks the assets origin works [^1] and predates
+  # the other tests here [^2] that cover the same in more depth
+  # e.g. "Check assets with an xls extension are served correctly".
+  #
+  # [^1]: https://github.com/alphagov/govuk-puppet/blob/32c1bbbb10067078c1406170666a135b4a10aaea/modules/router/templates/assets_origin.conf.erb#L76
+  # [^2]: https://github.com/alphagov/smokey/commit/91bb195fff35b3398a8f442ba489284964a9e0ea
+  #
   Scenario: Check assets are being served
     Given I am testing "assets"
     When I request "/__canary__"
@@ -15,18 +29,38 @@ Feature: Assets
     When I request "/assets/513a0efbed915d425e000002"
     And I should see "120613_Albania_Travel_Advice_WEB_Ed2_jpeg.jpg"
 
+  # TODO: REMOVE as this test is a duplicate.
+  #
+  # Duplicate of "Check assets with an xls extension are served correctly".
+  #
   Scenario: Check an asset can be served
     Given I am testing "assets"
     And I force a varnish cache miss
     When I request "/media/513a0efbed915d425e000002/120613_Albania_Travel_Advice_WEB_Ed2_jpeg.jpg"
     And I should get a content length of "212880"
 
+  # TODO: REMOVE as this test is a duplicate.
+  #
+  # Duplicate of "Check assets with an xls extension are served correctly".
+  #
+  # Note: this was originally testing a bug [^1][^2] but no longer:
+  # the object in S3 no longer has a content type set as described.
+  #
+  # Since this feature was at best an intermittent way of detecting
+  # the bug and since it's been broken for an unknown period, I think
+  # it's safe to remove it for now.
+  #
+  # [^1]: https://github.com/alphagov/smokey/commit/3725f865d22340698fad39ae7f3cca20d0d2ecfa
+  # [^2]: https://github.com/alphagov/govuk-puppet/commit/79a0bfec37d193c55c40787375a721408db606ce
+  #
   Scenario: Check assets with a docx extension are served correctly
     Given I am testing "assets"
     And I force a varnish cache miss
     When I request "/media/59f70d5640f0b66bbc806ed3/questionnaire-for-accommodation-providers-online-hotel-booking.docx"
     Then I should get a "Content-Type" header of "application/vnd.oasis.opendocument.text"
 
+  # TODO: RENAME to clarify this is testing that an asset can be served.
+  #
   Scenario: Check assets with an xls extension are served correctly
     Given I am testing "assets"
     And I force a varnish cache miss
--- a/features/benchmarking.feature
+++ b/features/benchmarking.feature
@@ -1,6 +1,15 @@
 Feature: Benchmarking
   Tests to check the loading times for various pages on GOV.UK.
 
+  # TODO: IMPROVE this test.
+  #
+  # Remove the timing step as it does not meet the eligibility
+  # criteria in docs/writing-tests.md.
+  #
+  # - Covers site-wide config: N (not applicable)
+  # - Targets data transfer: N (out-of-scope)
+  # - Second critical check: N (not applicable)
+  #
   @local-network
   Scenario: Check Bouncer application is up
     Given I am testing "bouncer" internally
@@ -10,6 +19,17 @@ Feature: Benchmarking
     And I should get a "Location" header of "https://www.gov.uk/government/organisations/attorney-generals-office"
     And the elapsed time should be less than 2 seconds
 
+  # TODO: REMOVE this test as it is a duplicate.
+  #
+  # Duplicate of "Check licence finder loads".
+  #
+  # The timing step does not meet the elgibility criteria
+  # in docs/writing-tests.md.
+  #
+  # - Covers site-wide config: N (not applicable)
+  # - Targets data transfer: N (out-of-scope)
+  # - Second critical check: N (not applicable)
+  #
   Scenario: Check the licence finder home page loads quickly
     Given I am benchmarking
     And I am testing through the full stack
@@ -17,14 +37,36 @@ Feature: Benchmarking
     When I visit "/licence-finder"
     Then the elapsed time should be less than 2 seconds
 
+  # TODO: REMOVE this test as it is a duplicate.
+  #
+  # Duplicate of "Check licensing app is present".
+  #
+  # The timing step does not meet the elgibility criteria
+  # in docs/writing-tests.md.
+  #
+  # - Covers site-wide config: N (not applicable)
+  # - Targets data transfer: N (out-of-scope)
+  # - Second critical check: N (not applicable)
+  #
   Scenario: Check requesting a PDF takes a reasonable amount of time
     Given I am testing "licensing" internally
       And I am benchmarking
       And I am testing through the full stack
       And I force a varnish cache miss
-    When I request "/apply-for-a-licence/forms/bury/test-licence/9999-7-1,0-1"
-    Then the elapsed time should be less than 10 seconds
+      When I request "/apply-for-a-licence/forms/bury/test-licence/9999-7-1,0-1"
+      Then the elapsed time should be less than 10 seconds
 
+  # TODO: REMOVE this test as it does not meet the eligibility
+  # criteria in docs/writing-tests.md.
+  #
+  # - Covers site-wide config: N (not applicable)
+  # - Targets data transfer: N (already covered, out-of-scope)
+  # - Second critical check: N (no feature tested)
+  #
+  # This page is actually rendered by Government Frontend. Data
+  # transfer for this app with Content Store is already covered
+  # by "Ensure static content is rendered".
+  #
   Scenario: Check whitehall-frontend pages load quickly
     Given I am benchmarking
     And I am testing through the full stack
@@ -32,6 +74,17 @@ Feature: Benchmarking
     When I visit "/government/how-government-works"
     Then the elapsed time should be less than 2 seconds
 
+  # TODO: REMOVE this test as it is a duplicate.
+  #
+  # Duplicate of "Check the whitehall world locations API returns data".
+  #
+  # The timing step does not meet the elgibility criteria
+  # in docs/writing-tests.md.
+  #
+  # - Covers site-wide config: N (not applicable)
+  # - Targets data transfer: N (out-of-scope)
+  # - Second critical check: N (not applicable)
+  #
   Scenario: Check the whitehall API loads quickly
     Given I am benchmarking
     And I am testing through the full stack
@@ -39,6 +92,17 @@ Feature: Benchmarking
     When I visit "/api/world-locations"
     Then the elapsed time should be less than 2 seconds
 
+  # TODO: REMOVE this test as it does not meet the eligibility
+  # criteria in docs/writing-tests.md.
+  #
+  # - Covers site-wide config: N (not applicable)
+  # - Targets data transfer: N (already covered, out-of-scope)
+  # - Second critical check: N (no feature tested)
+  #
+  # This page is rendered by Finder Frontend. Data transfer for
+  # this app with Content Store and Search API is already covered
+  # by e.g. "Check search results and analytics".
+  #
   Scenario: Check the research and statistics page loads quickly
     Given I am benchmarking
     And I am testing through the full stack
--- a/features/feedback.feature
+++ b/features/feedback.feature
@@ -5,15 +5,43 @@ Feature: Feedback
     Given I am testing through the full stack
     And I force a varnish cache miss
 
+  # TODO: RENAME to clarify this is testing data transfer with
+  # Content Store (as the prime example for this app).
+  #
   Scenario: Check the "Contact GOV.UK" page loads correctly
     When I visit "/contact/govuk"
     Then I should see "Contact GOV.UK"
 
+  # TODO: EXPORT this test as it does not meet the elgibility
+  # criteria in docs/writing-tests.md.
+  #
+  # - Covers site-wide config: N (not applicable)
+  # - Targets data transfer: N (already covered)
+  # - Second critical check: N (not tested in app)
+  #
+  # Data transfer with Content Store is already covered by
+  # "Check the "Contact GOV.UK" page loads correctly".
+  #
+  # Should be tested in Feedback [^1].
+  #
+  # [^1]: https://github.com/alphagov/feedback/blob/d2e8d8286fffec6b6093eecd5ed2d38f2c245e9e/app/models/contact_ticket.rb#L16
+  #
   Scenario: Check malicious code does not execute
     When I visit "/contact/govuk"
     And I input malicious code in the email field
     Then I see the code returned in the page
 
+  # TODO: REMOVE this test as it does not meet the eligibility
+  # criteria in docs/writing-tests.md.
+  #
+  # - Covers site-wide config: N (not applicable)
+  # - Targets data transfer: N (already covered)
+  # - Second critical check: N (arbitrary page)
+  #
+  # The URL is redirect and is rendered by Government Frontend.
+  # Data transfer for this app with Content Store is already
+  # covered by "Ensure static content is rendered".
+  #
   Scenario: Check the FoI page loads correctly
     When I visit "/contact/foi"
     Then I should see "How to make a freedom of information (FOI) request"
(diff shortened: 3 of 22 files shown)


KludgeKML (approved):
Looks good to me (having checked docs, licence-finder, smart answers - and manuals-frontend has be removed completely, so that's definitely fine).

ollietreend (approved):

richardTowers (commented):
Just putting a 👍🏻  rather than formally approving, because I haven't had time to look through all the comments and check they make sense.

I certainly agree with the spirit of this PR - we should be applying some pressure to reduce the number of browser based tests we do, otherwise they'll just grow and grow over time. From what I've seen, this is a thoughtful and well reasoned approach to doing that.

gclssvglx (approved):

docs/writing-tests.md
```diff
@@ -0,0 +1,60 @@
...
+
+Do not add tests that can be implemented in other ways. We want this to be the tip of the [testing pyramid](https://martinfowler.com/bliki/TestPyramid.html), not a [testing ice cream cone](http://saeedgatson.com/the-software-testing-ice-cream-cone/). Instead of adding a test here, you can:
+
+- Add or extend [health checks for application infrastructure](https://github.com/alphagov/govuk_app_config/blob/main/docs/healthchecks.md). For example, drafting a document in a publishing app involves connecting to infrastructure like a database or AWS S3.
+
+- Add [contract tests](https://docs.publishing.service.gov.uk/manual/pact-broker.html) to cover a chain of APIs. For example, subscribing to an email involves communication with Email Alert API and Content Store, which can be tested via GDS API Adapters.
+
+- Add unit or integration tests to cover in-app behaviour. For example, the way each document format is rendered can be tested in the frontend app that does the rendering.
+
```
theseanything, line 39: We have a few tests that try to measure latency - which is better served by monitoring application metrics rather than a test. Using a test is guaranteed to be flakey - as we never (and shouldn't) expect 100% of requests to meet our requirements. 

Wonder if we should add that here as guidance?

docs/writing-tests.md
```diff
@@ -0,0 +1,60 @@
...
+In contrast, the tests in this repo provide coverage at a higher level. We should be willing to page someone when tests are failing, once they are reliable enough.
+
+## Adding new tests
+
+We use feature files to describe single applications (e.g. [`whitehall`](https://github.com/alphagov/whitehall), [`frontend`](https://github.com/alphagov/frontend)), as well as [cross-application behaviour](https://github.com/alphagov/smokey/blob/main/features/gov_uk.feature).
+
+Try to use existing step definitions e.g. [`When I visit "/some/path"`](https://github.com/alphagov/smokey/blob/b656e65ef6853d65de3e24a7e884911054c8bf7d/features/step_definitions/smokey_steps.rb#L70-L72). If a more specific step is required, name the `_steps.rb` file after the feature file i.e. `myfeature_steps.rb`.
+
```
theseanything, line 47: We have a lot of tests that seem to be concerned with the implementation of GOV.UK, rather than just testing the user functionality (e.g. Check homepage is served by Varnish). Once we start testing configuration our test become more fragile and they aren't testing what we actually care about (i.e. that we can just load the homepage). For example, if we change the configuration but the feature still works when end up breaking tests.

Would be good if we had guidance that new tests are written to test user features, rather than configuration.

theseanything (approved):
Thanks Ben - haven't had a comprehensive review, but definitely agree with a lot of the changes. I think this is a step in the right direction to have better guidance and understand what the purpose of Smokey test for

injms (approved):

docs/writing-tests.md
```diff
@@ -0,0 +1,60 @@
...
+- Add [contract tests](https://docs.publishing.service.gov.uk/manual/pact-broker.html) to cover a chain of APIs. For example, subscribing to an email involves communication with Email Alert API and Content Store, which can be tested via GDS API Adapters.
+
+- Add unit or integration tests to cover in-app behaviour. For example, the way each document format is rendered can be tested in the frontend app that does the rendering.
+
+In contrast, the tests in this repo provide coverage at a higher level. We should be willing to page someone when tests are failing, once they are reliable enough.
+
+## Adding new tests
+
+We use feature files to describe single applications (e.g. [`whitehall`](https://github.com/alphagov/whitehall), [`frontend`](https://github.com/alphagov/frontend)), as well as [cross-application behaviour](https://github.com/alphagov/smokey/blob/main/features/gov_uk.feature).
```
kevindew, line 44: Is there a pattern for when you're testing how data flows between applications as to which application you test in? i.e. do you put it in the application you assert against? Only a quibble that if there is a pattern we might want to make that clearer.

kevindew (approved):
Looks great, thanks for all your work on this ⭐

docs/writing-tests.md
benthorner: Hmm...this is tricky one.

I don't think we should change the guidance as it stands until an alternative exists. I do agree that the mixture of user-focussed and non-user-focussed tests makes the overall approach to writing tests a bit confusing. One of the changes I plan to make after this PR is to reorganise some of the tests, so the "config" ones should be a bit more contained.

Having audited them all, I don't think the existing "config" tests are fragile e.g.

- "Check homepage is served by Varnish (CDN)" is unlikely to be affected by changes to the home page.
- "Check redirect from bare domain to www.gov.uk is working for HTTP" should also be robust longterm.

I think I get your point though: by writing tests that assert at a lower level, we set a precedent for tests that _are_ fragile, because they're too granular. It's worth noting this can happen just as much with (so called) user-focussed tests ([example](https://github.com/alphagov/smokey/pull/965/files#diff-d806eef09bf240d54b61cc2de907d6cb57806d48329fc29884f0c0cc5556d7f6R58)). In other words, I don't think the "config" tests represent a real problem in the way you describe.

However, in the longterm we should have a separate suite of tests for things like the CDN, where we fully test everything we care about - "vital" or not. But if we disallow new "config" tests now we effectively force a developer to create the alternative, which might happen, or the test might not get written if they think it's too difficult.

I'm seeing this PR as a step - a prerequisite, really - towards these kinds of changes.

What do you think? Did I understand you correctly?

docs/writing-tests.md
benthorner: Yes, we should clarify this. I've already written [the draft guidance for it](https://docs.google.com/document/d/1JvrurIoaYN_Ob_KlY7fecyvG2OGZoZpL0eApeZ5B-14/edit#heading=h.4ia795io3nzc) but have omitted it in this PR because I want to take a similar approach of applying it at the same time as adding it, which will be a lot easier after this change. Right now the tests are all over the place, which is consistent with the (lack of) specific guidance here.

Since we're discussing it, though:

> do you put it in the application you assert against?

Yes, this is what I mean by "target":

> “<app>.feature” should contain tests that target a feature of that app

I've used the word "target" in this guidance for app-to-app data transfer. Do you think it's clear enough, or would "assert about / against" be better?

JonathanHallam (approved):
Nice one, thanks for this Ben

benthorner:
Minor update: I've rebased against `main` so this can be merged. 

As part of the rebase I moved [two examples referring to manuals-frontend](https://github.com/alphagov/smokey/pull/964/files#diff-5ccb582a6dfcda27624ab1e4a55af91f1a9ebceee11c13322ccd731288b51813R43) (now rendered by government-frontend) and marked [an old manuals-frontend test](https://github.com/alphagov/smokey/pull/964/files#diff-e4b9d3e324849d61dd64492678aa2cad46223889dd727e4b581e81ca99be0f6cR35) for [removal](https://github.com/alphagov/smokey/pull/965/files#diff-e4b9d3e324849d61dd64492678aa2cad46223889dd727e4b581e81ca99be0f6cR45). Neither should be contentious.

I also noticed [one of the "EXPORT" tags in `assets.feature` should have been "REMOVE"](https://github.com/alphagov/smokey/pull/965/files#diff-a7610d8b814030196af66c95d5a1fd9f0982cbe1bdd1a8810ca0e4a5f14b6467R6) and have updated it. The explanation is unchanged so this shouldn't be contentious either.

features/assets.feature
```diff
@@ -15,18 +29,38 @@ Feature: Assets
     When I request "/assets/513a0efbed915d425e000002"
     And I should see "120613_Albania_Travel_Advice_WEB_Ed2_jpeg.jpg"
 
+  # TODO: REMOVE as this test is a duplicate.
+  #
+  # Duplicate of "Check assets with an xls extension are served correctly".
```
maxgds, line 34: There's an argument that this isn't a duplicate - are we confident that anything affecting asset type X will always affect asset type Y?

features/assets.feature
```diff
@@ -15,18 +29,38 @@ Feature: Assets
     When I request "/assets/513a0efbed915d425e000002"
     And I should see "120613_Albania_Travel_Advice_WEB_Ed2_jpeg.jpg"
```
maxgds, line 30: Looks like a lot of these scenarios could do with syntax being tidied up - the assertions should being with "then", possibly followed by "and"s.

features/benchmarking.feature
```diff
@@ -1,6 +1,15 @@
 Feature: Benchmarking
   Tests to check the loading times for various pages on GOV.UK.
 
+  # TODO: IMPROVE this test.
+  #
```
maxgds, line 5: Improve or remove? If you want to drop the timing then it no longer meets the stated reason for being there in the feature description.

features/feedback.feature
```diff
@@ -5,15 +5,43 @@ Feature: Feedback
...
+  # criteria in docs/writing-tests.md.
+  #
+  # - Covers site-wide config: N (not applicable)
+  # - Targets data transfer: N (already covered)
+  # - Second critical check: N (not tested in app)
+  #
+  # Data transfer with Content Store is already covered by
+  # "Check the "Contact GOV.UK" page loads correctly".
+  #
+  # Should be tested in Feedback [^1].
```
maxgds, line 25: I suggest we change this instead of removing - feedback should test the contact form directly, but there is a path from the component at the bottom of any given page behind the "report a problem" button that uses a combination of frontend app X interfacing with the feedback form.

features/gov_uk.feature
```diff
@@ -6,25 +6,91 @@ Feature: Core GOV.UK behaviour
...
+  #
+  # - Covers site-wide config: N (edge case)
+  # - Targets data transfer: N (not applicable)
+  # - Second critical check: N (not applicable)
+  #
+  # The low impact and likelihood of this breaking compared to
+  # any other CDN feature means the risk of removal is low.
+  #
+  # No evidence this is a vital feature [^1].
+  #
```
maxgds, line 39: I think things like email clients can be a bit naive about auto-linking URLs and sometimes mistakenly include full stops. I think there's a risk of people thinking this isn't vital and it breaking and departmental coms leading to users not seeing pages. Inclined to keep this.

features/assets.feature
benthorner: Good spot. I've found lots of these little issues as I've been going through the tests. They're out-of-scope for this PR, but I'm in the area in one of the follow-up PRs I've got planned, then I'll look at this.

features/benchmarking.feature
benthorner: It's a bit of both really: the performance steps don't meet the eligibility criteria, but the rest of the scenario is OK and is the only scenario we have that's testing the Bouncer app, so we should try to keep it.

Unlike all the other scenarios in here, which are covered elsewhere, my intention is to just remove the offending steps and move / rename the scenario so its value is clear - as the sole test Bouncer is working.

Does that help? I can tweak the comment but want to check I understand first.

maxgds: Yeah, that makes sense.

features/feedback.feature
maxgds: 1, yes. 2 though, I THINK the component builds its own form at the bottom of every page (though it might be magic ajax style shenanigans pulling through the Feedback form). Either way, it posts to the feedback app. I think there's a chance this vulnerability may be there as an independent thing, but more importantly we need to know that a page rendered by random app X can send content successfully to the feedback app, I think?

features/gov_uk.feature
benthorner: Thanks for challenging this - I struggled with these so appreciate the input.

It sounds like this is different to the other scenarios, which are focussed on human error: you're saying this could actually manifest through software, which does add weight to it being useful.

My worry with this is that it sets a precedent for having other edge-case tests: while a full stop may be more of an issue, I suspect it still accounts for a small minority of our daily requests.

Here are some options:

1. I can add a commit to keep this test until we have data to justify it's not vital e.g. if it only applies to .00000001% of our daily requests, I think that is a strong justification that this test has very low value.

2. We can continue with the removal. There's always a risk with removing a test but we can still agree it's worth the risk, even with this new knowledge. I still lean towards removal, but it's ultimately subjective.

With that in mind, which option do you think I should go with?

features/assets.feature
benthorner: Yes, there's nothing special about serving image files vs other files.

Asset Manager does have [a bit of special code for image files that don't exist](https://github.com/alphagov/asset-manager/blob/e8b25f21a72948117c7dcb084491af26b949805a/app/controllers/whitehall_media_controller.rb#L27) (unsurprisingly, it's to accommodate Whitehall), but that's not relevant for the test here, where we're testing files that do exist.

features/feedback.feature
benthorner: Here we go: ["report a problem" is rendered as part of the component](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/feedback/_problem_form.html.erb).

We do have a data transfer test for one of the other API calls to feedback (below). AJAX features are admittedly a bit harder to cover elsewhere because we can't use our normal Ruby test approaches, so I think there is an argument for having a new test to cover "report a problem" specifically talking to feedback.

However, I want to focus on the tests we can remove (or export) first. The check for malicious content is part of the feedback app, so we won't lose anything if we rewrite this test there - if the protection is already being shared with the "report a problem" API then it will still be covered, implicitly or explicitly.

I think that covers everything - quite a complex discussion.

The key question: is there a reason we shouldn't EXPORT this test to feedback?

Happy to chat if there's more to discuss - might be easier at this point.

maxgds: It does make sense to export that exact test. I think there is a gap in here that could potentially be another "TO DO".

features/gov_uk.feature
benthorner: Fair enough.

@kevindew @richardTowers could one of you give us a steer on whether this is an acceptable risk?

It's just one test, so I don't feel strongly - we can keep it if you're not sure.

docs/writing-tests.md
kevindew: Great stuff. The guidance sounds good. I think target sounds reasonable. Though I interpret it slightly different to assert. I may be caught on an uncommon edge-case though. For example if we're targetting an email-alert-api feature, but we assert on email-alert-frontend I'd assume we'd have an email-alert-api feature to test against?

docs/writing-tests.md
theseanything: Yeah, I guess just adding guidance here could help prevent regression and someone adding a new test for latency.

docs/writing-tests.md
kevindew: Yup makes sense. I think I'm probably a little too much in edge case land considering how we might be targeting behaviour of an app that is different to what we're asserting against. 

Vague seems reasonable, if possible a slight rephrase to indicate app asserting against could be somewhat beneficial but not worth getting stuck on.

benthorner: Actually, I just found an example:

> “[Check visiting a draft page requires a signon session](https://github.com/alphagov/smokey/blob/aaa0522666d49db116b939276f0e0969357e3ed4/features/draft_environment.feature#L9)” should be located in “authenticating_proxy.feature”.

Here we're asserting against government-frontend but actually targeting a feature of the proxy. Another reason to keep it vague is that it would make some of the existing tests appear ineligible, where the transfer of data is only being tested implicitly ([example](https://github.com/alphagov/smokey/blob/68292a3a54475f093f1ff158367cbf5cb94e4e31/features/whitehall.feature#L21)). While we should improve these, I think that should be done as part of a future iteration.

Short version: best to leave it as-is I think - the vagueness is deliberate.

features/gov_uk.feature
kevindew: Sure, I've just had a look into it.

I can see from [Kibana](https://kibana.logit.io/s/2dd89c13-a0ed-4743-9440-825e2e52329e/goto/ff16963394a0b4f914b23737de857e48) that we've had ~30,000 of requests that get this redirect compared to ~177,000,000 requests - so perhaps 0.02% of requests that aren't handled at CDN (which still seems sufficiently high that I fear I've made a mistake).

I think it's an acceptable risk. Ideally this would be tested with the VCL file that configures this: https://github.com/alphagov/govuk-puppet/blob/32c1bbbb10067078c1406170666a135b4a10aaea/modules/varnish/templates/default.vcl.erb 

So I'd think it'd be valuable here if we decided that this is the suite to test varnish and we had a varnish.feature which tests each thing added. As it's currently a hodge-podge I think a better precedent would be to remove them, rather than adding a bunch of missing varnish tests (for consistency). It doesn't seem ideal to have a situation where we selectively test bits of VCL.

docs/writing-tests.md
benthorner: Sorry, I don't think I was clear in my comment: we already have a "bad example" about that sort of test, which I think of as guidance. Do you think it's worth being even more explicit about performance tests?

docs/writing-tests.md
theseanything: Ahh right, maybe the name for that example I gave was confusing - it actually tests whether the pages is cached (which is fine), rather than checking that its being served explicitly by Varnish. That would have created a fragile test - as we could change the caching layer to another service other than varnish, and it would have broken the test, but we'd still be caching the request (the thing we care about).

I can't find another test that tests configuration like this with a quick glance over. So don't want this comment to block the PR.

features/gov_uk.feature
benthorner: > so perhaps 0.02% of requests that aren't handled at CDN

Ooh thanks for checking. I'm surprised I have access to Production Kibana 🤔.

> which still seems sufficiently high that I fear I've made a mistake

Yes, I think it's even lower: looking through the logs I can see we're also catching query params and invalid URLs that happen to end in a `.`. We're likely seeing a 301 for the bad URLs due to `http => https`.

I'm not sure how to do better, but 0.02% is low. The scenario of a bad link in an email is still possible, but we know most of our traffic doesn't come from these - at least, not from the emails we send ourselves.

I'm going to resolve this one as we've agreed it's an acceptable risk.

1pretz1 (commented):
Looks good Ben - cheers for putting this together ⭐ (sorry the comment is post merge 😄)

PR #967, opened 2022-06-07
Simplify test setup - bypass caching by default

benthorner:
This continues work started in [^1].

Most of the tests are trying to visit the most recent
(uncached) version of a page on GOV.UK. Only a handful
of tests are checking actual caching behaviour, so it
makes sense to bypass caching by default. Bypassing the
CDN only makes sense if we're visiting GOV.UK - versus
some internal app domain - so I've made this combo the
new default for all the tests.

[^1]: https://github.com/alphagov/smokey/pull/886

## Testing


KludgeKML (approved):
All seems to make sense to me.

benthorner:
Adding "DNM" as this conflicts with https://github.com/alphagov/smokey/pull/965, which takes precedence over this.

benthorner:
Rebased and retested - all still works ✅.

PR #972, opened 2022-06-15
Remove and rename scenarios as per the TODOs

benthorner:

## Testing


features/benchmarking.feature
```diff
@@ -1,111 +1,9 @@
...
-  # Remove the timing step as it does not meet the eligibility
-  # criteria in docs/writing-tests.md.
-  #
-  # - Covers site-wide config: N (not applicable)
-  # - Targets data transfer: N (out-of-scope)
-  # - Second critical check: N (not applicable)
-  #
   @local-network
-  Scenario: Check Bouncer application is up
+  Scenario: Check redirects work for transitioned sites
```
ollietreend, line 5: Does this test still belong in `benchmarking.feature`? Is there a more appropriate location for it now, given that it's no longer about benchmarking.

benthorner: Indeed it does not - it should go in something like `bouncer.feature`. 

There are quite a few scenarios that are poorly located, which I'm going to address with a little extra guidance that builds #965. You can see the draft guidance [here](https://docs.google.com/document/d/1JvrurIoaYN_Ob_KlY7fecyvG2OGZoZpL0eApeZ5B-14/edit#heading=h.4ia795io3nzc) (starting "Scenarios should be organised").

For now I'd like to keep everything where it is and explain all the moves in one go, along with the necessary changes to Puppet to maintain alerts. I appreciate it's a little bit confusing, but it won't be for very long.

Is that OK-ish?

ollietreend (approved):

PR #973, opened 2022-06-15
Rename scenarios to be more consistent

benthorner:
https://trello.com/c/fRihmgQ8/266-add-top-level-guidance-about-where-we-write-tests-on-govuk

Another follow-up PR to #965. I've gone a bit further and looked at all the 
scenarios to see where the names could be better. Please see the commits 
for more details. After this we will have a minimal set of TODOs remaining 
and can move on to other work e.g. clarifying which files tests should be in.

If you're wondering about some of the claims of which app talks to which 
other app, [the original Smokey audit](https://docs.google.com/spreadsheets/d/1i2k4WexywoJurdFddOxiNICN-Elv1D5Mb2sdaCjF1t0/edit?disco=AAAASMdZEUQ) has links to supporting evidence for 
most of the tests ([example](https://docs.google.com/spreadsheets/d/1i2k4WexywoJurdFddOxiNICN-Elv1D5Mb2sdaCjF1t0/edit?disco=AAAASMdZEUQ)).

Tested and works on [Integration Jenkins](https://deploy.integration.publishing.service.gov.uk/job/Smokey/40478/console) ✅*.

*The failures are also on `main` and unrelated to these changes.

PR #976, opened 2022-06-16
Reorganise scenarios based on additional guidance

benthorner:

This was meant to go in #965 but would have made the PR too big to 
review. Separating it here also makes it practical to show by doing and 
move scenarios so they comply with the guidance.

The number of feature files has increased, but there were already lots 
of them. In future work we could look at putting all the `<app>` features 
in a subdirectory, now they're in an actual pattern.

Some of the scenario names and tags had to change a bit - see the 
commit messages for more details.

I've not made any changes to the step definition files. The steps are 
already disorganised with respect to the features and it's not worth 
bloating the commits trying to keep them in check.

**TODO**: this will need to be merged in tandem with a PR in Puppet 
to adjust the Smokey Loop alerts, which are coupled to the feature 
files we have here ([example](https://github.com/alphagov/govuk-puppet/pull/11713)).

## Review

- Read the new guidance ([rendered version](https://github.com/alphagov/smokey/blob/organising-tests/docs/writing-tests.md#adding-new-tests))

- Do the following `<layer>.feature` files make sense?

  - origin.feature
  - assets.feature
  - mirror.feature
  - cdn.feature

- Do the `<app>.feature` files still make sense?

- Does [the new file structure](https://github.com/alphagov/smokey/tree/7ddeb78cab1fe160de5455e1ea03f93cd3e3ffaa/features) make sense?

## Testing

Tested and works on [Integration Jenkins](https://deploy.integration.publishing.service.gov.uk/job/Smokey/40563/console) ✅*.

*The test failures are unchanged from `main`.


features/collections.feature
```diff
@@ -19,3 +19,15 @@ Feature: Collections
...
+    When I visit "/government/organisations"
+    And I confirm it is rendered by "collections"
+    And I click to report a problem with the page
+    Then I see the report a problem form
+    When I close the open feedback form
+    And I click to say the page is not useful
+    Then I see the email survey signup form
+    When I close the open feedback form
+    And I click to say the page is useful
+    Then I see the feedback confirmation message
```
ollietreend, line 33: This repetition feels a bit yuck, but I'm not sure what (if anything?) can be done about it. You've followed your own rules of splitting tests by application.

It just makes me wonder if there's a less repetitive way for us to test features that span multiple applications because they exist in the global GOV.UK page layout. I suppose this could be considered to be similar to `Scenario: Check the crown logo links to GOV.UK homepage` in that regard.

ollietreend (commented):
@benthorner I think this looks good 👍🏻

The guidance reads well, and I think the `<layer>.feature` and `<app>.feature` naming strategy makes sense to me. However I do think these `.feature` files would benefit from being moved into `infrastructure/` and `apps/` subdirectories to make them easier to differentiate.

I've added a comment about something which 'feels' bad, but I understand why you've done it and don't have any better suggestions.

features/collections.feature
benthorner: I agree it's yuck - more visibly yuck, at least.

First I need to make a correction: https://github.com/alphagov/smokey/pull/973/commits/0091108011cb5b95440798f975af57e04194fc3b was a poor change because this is not secondary, nor is it about site-wide config; it really is targeting communication with Static and I should revert that change.

Based on [the original commit](https://github.com/alphagov/smokey/commit/ca008962fe89e1b62a5df97a47b0db020670a542), the need for this test is likely to go away with Static: if we still really want to test this, it will be possible to do via the component and / or individual apps.

I'd like to explore a partial deletion option: is the data transfer with Static for all of these apps actually a critical part of GOV.UK as a whole? I don't think this is true for something like info-frontend, for example.

I'm wondering if we could cut this down to just:

- government-frontend
- frontend
- finder-frontend
- collections
- smartanswers

That would cut out half the repetition. Does that change anything for you? This is one of those classic testing problems where we've DRYed-up the code but still need tests for each place it's used.

benthorner:
> these .feature files would benefit from being moved into infrastructure/ and apps/ subdirectories

@ollietreend this was actually my original plan, but I thought it would be a step too far for this PR, noting that it also requires extra changes here and Puppet to support subdirectories for Smokey Loop.

However, on reflection I think it's easier to argue for the changes here if the end result is tidier, so I've tweaked how Smokey Loop runs in https://github.com/alphagov/smokey/pull/976/commits/c537962606510e5dba8ab3340f6620802dd981bf and moved all the app features in https://github.com/alphagov/smokey/pull/976/commits/7ddeb78cab1fe160de5455e1ea03f93cd3e3ffaa.

I don't think it's worth having another sub-directory for "layers" or "infrastructure" as the pattern is fairly weak and doesn't account for outliers like `data_gov_uk.feature` and `licensing.feature` (out-of-scope).

Note: I've confirmed the changes in the last commit will work.

| After deploying Puppet | After deploying Smokey (as well) |
| - | - |
| Screenshot 2022-06-20 at 11 51 17 | Screenshot 2022-06-20 at 11 57 06 |

docs/writing-tests.md
```diff
@@ -41,7 +41,19 @@ In contrast, the tests in this repo provide coverage at a higher level. We shoul
 
 ## Adding new tests
 
-We use feature files to describe single applications (e.g. [`whitehall`](https://github.com/alphagov/whitehall), [`frontend`](https://github.com/alphagov/frontend)), as well as [cross-application behaviour](https://github.com/alphagov/smokey/blob/main/features/gov_uk.feature).
+Scenarios should be organised by the "component" they are testing:
+
+- **`<app>.feature`** should contain tests that target a feature of that app e.g.
```
ollietreend, line 46: Following 7ddeb78cab1fe160de5455e1ea03f93cd3e3ffaa where you moved the app-focussed features into an `app/` subdirectory, it'd be good to update this guidance to match.

And then as you said in https://github.com/alphagov/smokey/pull/976#issuecomment-1160318013, it sounds like you're suggesting the `<layer>.feature` files remain as top level features.

ollietreend (approved):
This looks good to me @benthorner. Great work putting this together! ✨🚀

features/collections.feature
kevindew: Sorry for being the one who introduced this pain point - I recall adding to Smokey being pretty much a last resort effort to add testing for something that wasn't testable in app. I'd agree it's not ideal to be a Smokey thing and would be much better served in the app, with Static being the pain point. I'm not quite sure why it can't be pretty much all handled in testing of the component gem but it's clearly way more complex than that now.

I guess the first thing I'd suggest for this is shall we explore making a step that can encapsulate the duplication. e.g:

```
  Scenario: Check the feedback component loads
    When I visit "/government/organisations"
    And I confirm it is rendered by "collections"
    Then I can operate with the feedback component
```

I think that'd make it much less painful to test.

> I'd like to explore a partial deletion option: is the data transfer with Static for all of these apps actually a critical part of GOV.UK as a whole? I don't think this is true for something like info-frontend, for example.

You're right. info-frontend get's so little love that it wouldn't be a biggie. Though it feels like for almost everything else we probably do care somewhat. E.g. email-alert-frontend, you can't sign in we kinda want people to be able to feedback, service-manual-frontend, 🤷. I worry though that partial deletion leaves it as rather ambiguous.

kevindew (approved):
Thanks a lot for this. Looks like great work.

I know you're keen to merge this so I'm happy to approve though I think it'd be good to still do some more exploring of dealing with the feedback feature. As discussed on Slack, I'm not totally sold on the "EXPORT" tag as I don't think it's easily exportable.

My current thinking is that all apps (now) embed the feedback component in the same way. This is done via [static calling the layout_for_public component](https://github.com/alphagov/static/blob/8bd186900d8f20a19c5354081294d95cf7e2ed09/app/views/root/_gem_base.html.erb#L43) which has it in. There is some magic that goes on with slimmer then trying to fix static's rendering of it. So I wonder if this is actually a feature of the static app and only needs testing on one page?

KludgeKML (approved):
Looks okay

benthorner:
@kevindew thanks for approving! Deploying this now will indeed help as it will allow Sentry to fill up with data for the new features (issues in Sentry), which I will then be able to use to update [the patterns I'm monitoring](https://github.com/alphagov/smokey/issues) (before I leave).

Thanks to @KludgeKML for speedy help with a speedy deployment 🌟.

Regarding feedback, I agree there's more to discuss and we should at least cut it down to one feature if you think that's reasonable. The question is still how to justify it's presence and how to document it if not.

I've made a note of this and we can revisit when I'm back next week.

PR #978, opened 2022-06-17
Improve test for visiting a draft asset

benthorner:
Previously we would try to view a non-existent asset, which didn't
matter as we were only checking the redirect to Signon, which works
irrespective of the asset existing [^1].

Now we test the full journey of viewing a draft asset*. I've chosen
to use an image because this will preview in the headless browser,
which is easier to check and avoids an actual download.

*Note: the asset is published but, like published documents, also
exists in draft as well.

Tested and works on [Integration Jenkins](https://deploy.integration.publishing.service.gov.uk/job/Smokey/40521/console) ✅.

[^1]: https://github.com/alphagov/asset-manager/blob/1e88cd5b71b8ac6f73a079176b9f74c57b608d1c/app/controllers/media_controller.rb#L54

### DIFF
--- a/features/assets.feature
+++ b/features/assets.feature
@@ -15,7 +15,9 @@ Feature: Assets
     When I request "/media/580768d940f0b64fbe000022/Target_incomes_calculator.xls"
     Then I should get a "Content-Type" header of "application/vnd.ms-excel"
 
-  Scenario: Check draft assets require authentication
+  Scenario: Check a draft asset can be served
     Given I am testing "draft-assets"
-    When I visit "/media/123/filename.extension"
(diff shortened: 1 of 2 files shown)


features/assets.feature
```diff
@@ -18,5 +18,7 @@ Feature: Assets
   @app-authenticating-proxy
   Scenario: Check draft assets require authentication
```
1pretz1, line 19: Should we match the naming convention of the scenario above as we're now checking we can preview the asset?

For example, 'Check a draft asset can be served'.

1pretz1 (approved):
Approving but with one minor, non-blocking comment.

features/assets.feature
benthorner: Good point. I've amended with that 👍.

PR #982, opened 2022-06-21
Set cache bypass to false on failing test

chao-xian:
Since we [bypass the cache by default](https://github.com/alphagov/smokey/pull/967), this CKAN search test was broken as it did not allow the `cache_bust` url param through.

## Testing

[This build](https://deploy.blue.production.govuk.digital/job/Smokey/11834/console) passes the [CKAN related failure](https://deploy.blue.production.govuk.digital/job/Smokey/11832/console ).

### DIFF
--- a/features/data_gov_uk.feature
+++ b/features/data_gov_uk.feature
@@ -28,6 +28,7 @@ Feature: Data.gov.uk
   @notintegration @notstaging
   Scenario: Check CKAN action api's search works
     Given I am testing "ckan"
+    And I try not to bypass caching
     When I request "/api/action/package_search?q=data"
     Then I should get a 200 status code
     And JSON is returned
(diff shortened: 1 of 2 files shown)


features/data_gov_uk.feature
```diff
@@ -28,6 +28,7 @@ Feature: Data.gov.uk
   @notintegration @notstaging
   Scenario: Check CKAN action api's search works
     Given I am testing "ckan"
+    And I do not bypass caching
```
benthorner, line 31:
```suggestion
    And I do not try to bypass caching
```

Optional: I think this is a slightly clearer as it hints at the mechanism in use; we're not sending some magical "do / don't cache signal", it's just a query param tacked onto the URL.

benthorner (approved):
Thanks for taking the time to fix this 💯.

Given the scope of the original change, I could have done more and checked the new code defaults also worked properly in Production.

I was also confused why we didn't see this problem earlier. Turns out we're missing some alerts. Could you take a look at https://github.com/alphagov/govuk-puppet/pull/11726?

features/data_gov_uk.feature
chao-xian: Nice suggestion!

PR #985, opened 2022-06-27
Tag asset manager in relevant whitehall tests

BeckaL:
As part of enabling CD for asset manager, this PR tags a few Whitehall tests that interact with asset manager so that we can ensure they are run as part of asset manager's deploy.

Tested on integration: https://deploy.integration.publishing.service.gov.uk/job/Smokey/40649/
(currently failing for unrelated reasons that can be seen on other earlier builds)

Trello: https://trello.com/c/Sj4UTatV/126-enable-continuous-deployment-for-asset-manager

## Testing

### DIFF
--- a/features/apps/whitehall.feature
+++ b/features/apps/whitehall.feature
@@ -4,11 +4,13 @@ Feature: Whitehall
     When I request "/government/ministers"
     Then I should see "Ministers by department"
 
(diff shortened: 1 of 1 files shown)


JonathanHallam (approved):
brill, cheers :)

PR #988, opened 2022-06-28
Remove excess feedback component scenarios

benthorner:

In response to [^1].

A single instance of this test is sufficient to cover the feedback
component working in general [^2].

Note: the government-frontend scenario was missed in the first pass
to DRY-up these scenarios for each app [^3].

[^1]: https://github.com/alphagov/smokey/pull/976#discussion_r900292101
[^2]: https://github.com/alphagov/smokey/pull/976#pullrequestreview-1015551923
[^3]: https://github.com/alphagov/smokey/commit/df3620d341848fcd2625439fc3b0681ce5dd8e86

## Testing


benthorner:
@kevindew here's a second pass at consolidating these tests based on our discussion so far. The next step would be to resolve the future of the single test that remains, but this PR represents a net improvement.

Before we proceed, I'd like to understand a bit more about why these tests were added in the first place - to validate they really are redundant. I can see [some fixes were made](https://github.com/alphagov/smokey/pull/837) but I can't see corresponding PRs to explain why this feature was so brittle across apps. I'll try a hunch, though: we were consolidating all layout components to be entirely served via static, which meant unpicking bespoke uses of components like this one. Was it something like that?

If these really are "short-term regression tests" then I agree we need at most one of them.

kevindew:
> @kevindew here's a second pass at consolidating these tests based on our discussion so far. The next step would be to resolve the future of the single test that remains, but this PR represents a net improvement.
> 
> Before we proceed, I'd like to understand a bit more about why these tests were added in the first place - to validate they really are redundant. I can see [some fixes were made](https://github.com/alphagov/smokey/pull/837) but I can't see corresponding PRs to explain why this feature was so brittle across apps. I'll try a hunch, though: we were consolidating all layout components to be entirely served via static, which meant unpicking bespoke uses of components like this one. Was it something like that?

Yes, that is correct. 

As I understand. Prior to switching to the gem layout apps used to embed the [feedback component themselves](https://github.com/alphagov/government-frontend/commit/f834104699a722eb6f7a98896d5b4fd94b2ec040). At this point in time the component could (and should) have been tested in app tests.

As apps started to swap to using the static provided gem_layout there was a period of time where we wanted consistent feedback behaviour across the site but the implementations differed. Hence all the smoke tests for each app. It probably didn't need them all at the time given some were still embedding the component directly - it seemed to offer insurance for when they were migrated. 

Now that all apps are using the gem_layout they should exhibit the same behaviour and have the same dependencies on static/slimmer/etc

features/apps/static.feature
```diff
@@ -0,0 +1,32 @@
...
+  # - Targets data transfer: N (not applicable)
+  # - Second critical check: N (not tested in app)
+  #
+  # Should be tested in Static [^1].
+  #
+  # [^1]: https://github.com/alphagov/static/blob/9f835b87a3aa6ed867b98b2f4534dd64c9510626/app/views/root/_gem_base.html.erb#L43
+  #
+  Scenario: Check the crown logo links to GOV.UK homepage
+    When I visit "/"
+    Then the logo should link to the homepage
```
kevindew, line 16: Can we consider deleting this test? It seems quite arbitrary. I notice the feature is tested in govuk_publishing_components: https://github.com/alphagov/govuk_publishing_components/blob/efea45dcbad13766ba73b028cb6cc3a2f99ce34c/spec/components/layout_super_navigation_header_spec.rb#L89-L93

So if we were to keep it, it'd probably count as a second critical check - but I don't think it's needed.

features/apps/static.feature
```diff
@@ -0,0 +1,41 @@
...
+  # - Targets data transfer: N (not applicable)
+  # - Second critical check: N (not tested in app)
+  #
+  # Partially tested in Slimmer [^1] but should be tested in
+  # Static [^2], although this test could still add value as
+  # a second critical check these two codebases can integrate.
+  #
+  # [^1]: https://github.com/alphagov/slimmer/blob/82144be372aa4afc1f8ff30917c96c2ee3a47ed8/test/processors/feedback_url_swapper_test.rb
+  # [^2]: https://github.com/alphagov/static/blob/9f835b87a3aa6ed867b98b2f4534dd64c9510626/app/views/root/_gem_base.html.erb#L43
+  #
```
kevindew, line 31: Sounds good 👍

features/apps/static.feature
kevindew: Ah, are you meaning that this is a test that static renders the gem_layout? 

It seems a pretty tenuous thing to test as we're basically just testing that a [variable set in static](https://github.com/alphagov/static/blob/577fd669535a8f5e44e26c2764e54a63ff51b793/app/views/root/_gem_base.html.erb#L4) gets past through a couple of layers of components (layout_for_public -> layout_super_navigation_header). 

I'd suggest it can be removed as global layout is covered implicitly in the [cookie banner](https://github.com/alphagov/smokey/blob/52f3f0cd73d56f9c7007b0c026109c7d568948db/features/apps/frontend.feature#L12-L15).

PR #989, opened 2022-06-28
Add Sentry tag for failures due to JS errors

benthorner:
We're getting a lot of these at the moment. This will help make it
easier to filter them out so we can see other types of flakiness.

Tested with a deployment of Smokey Loop in Integration.


gclssvglx (approved):
Great stuff - thanks @benthorner

PR #991, opened 2022-06-29
Remove Feedback export feature and corresponding steps

kevindew:
This can be removed now that the feature has been imported into Feedback
[1].

Note: I'm not able to remove the step definition for "I see the code
returned in the page" as this is still used by finder-frontend [2].
However once this PR is merged I should be able to add that onto [3].

[1]: https://github.com/alphagov/feedback/pull/1454
[2]: https://github.com/alphagov/smokey/blob/7553a432259d785fb6d4fb81cfd4fe3572185ad3/features/apps/finder_frontend.feature#L77
[3]: https://github.com/alphagov/smokey/pull/990

## Testing

### DIFF
--- a/features/apps/feedback.feature
+++ b/features/apps/feedback.feature
@@ -5,25 +5,6 @@ Feature: Feedback
     When I visit "/contact/govuk"
     Then I should see "Contact GOV.UK"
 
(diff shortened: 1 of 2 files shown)


benthorner (approved):
🎉

PR #992, opened 2022-06-29
Remove export get-involved features and corresponding steps 

kevindew:
Tests for the get-involved page have been written in government-frontend
now [1], so we can now remove these features that don't meet the Smokey
eligibility criteria.

[1]: https://github.com/alphagov/government-frontend/pull/2473

## Testing


benthorner (approved):
👍

(I agree about the step definition, although it's not alone in its fragility!)

PR #996, opened 2022-07-11
Revert "Bump selenium-devtools from 0.101.0 to 0.103.0"

kevindew:
Reverts alphagov/smokey#983

Smokey has been broken since this merged so trying to revert this seems sensible.


JonathanHallam (approved):
thanks :)

PR #1003, opened 2022-07-22
Update Dockerfile to use new GOV.UK Ruby base images

samsimpson1:
Updates the Dockerfile to use the new [GOV.UK Ruby base images](https://github.com/alphagov/govuk-ruby-images). Also updates the .ruby-version file to not specify a patch version, to allow for easier Ruby patching.

Context: https://trello.com/c/Zy0fd25w/970-use-base-builder-images-in-all-the-app-dockerfiles

## Testing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,15 @@
-ARG base_image=ruby:2.7.6
+ARG base_image=ghcr.io/alphagov/govuk-ruby-base:2.7.6
 FROM ${base_image}
-RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential \
-  libpq-dev libxml2-dev libxslt1-dev dumb-init default-jre
-
-ENV APP_HOME /smokey
-RUN mkdir $APP_HOME
 
 ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE 1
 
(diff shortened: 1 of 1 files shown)


MuriloDalRi:
@samsimpson1 is this still needed?

Dockerfile
```diff
@@ -1,15 +1,15 @@
...
-  libpq-dev libxml2-dev libxslt1-dev dumb-init default-jre
-
-ENV APP_HOME /smokey
-RUN mkdir $APP_HOME
 
 ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE 1
 
-# Install Chromium and ChromiumDriver
-RUN apt-get update -qq && apt-get install -y chromium chromium-driver && apt-get clean
+# Install dependencies, Chrimium and ChromiumDriver
```
ChrisBAshton, line 6: Typo:

```suggestion
# Install dependencies, Chromium and ChromiumDriver
```

But also I don't see much value in keeping the comment, so:

```suggestion
```

ChrisBAshton (commented):
The changes LGTM, @samsimpson1  - if you rebase against `main` you should trigger a build of the new CI for Smokey, which would also give some confidence for merging this change.

PR #1005, opened 2022-07-27
Remove TODO from Static feature test

robinjam:
[Trello card](https://trello.com/c/XDLC291t/2948-write-test-for-static-to-go-hand-in-hand-with-smokey-3)

This depends on another PR in Static[^1], which must be merged first.

This test has been reproduced in Static itself, but should be preserved here as a critical second check that the Static and Slimmer codebases can correctly integrate.

## Testing

[^1]: https://github.com/alphagov/static/pull/2835

### DIFF
--- a/features/apps/static.feature
+++ b/features/apps/static.feature
@@ -1,19 +1,5 @@
 Feature: Static
 
-  # TODO: EXPORT this test as it does not meet the eligibility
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
Looks good, pending merge of the other one 👍 thanks!

PR #1027, opened 2022-09-05
Add spec to handle Reshuffle Mode

chao-xian:
When we are in Reshuffle Mode, the Ministers page has different content and the smoke test for it fails.

## Testing


BeckaL (approved):
👍

PR #1029, opened 2022-09-21
Add continuous integration for Smokey itself

robinjam:
[Trello card](https://trello.com/c/FPbuquwu/2953-investigate-setting-up-ci-for-smokey-5)

This adds a quick ~sanity~ sense check when a branch is pushed or a PR is raised to ensure that Smokey isn't fundamentally broken (which can happen when e.g. Dependabot PR with incompatible changes is merged[^1]).

This has been tested on a separate branch to ensure that the test fails when the broken change from the PR above is merged[^2].

I still have some open questions about this PR, could the reviewer please let me know what they think:

- I'm not sure what name the workflow and job should have. Is "CI" appropriate? Maybe "~Sanity~ Sense check" would be more appropriate? Or something else?
- The test only fails for the above-mentioned Dependabot PR when the `RATE_LIMIT_TOKEN` environment variable is set (to any value), presumably because of the conditional [here](https://github.com/alphagov/smokey/blob/02784abbbcefe9e7366cf7b93d5f1937bdaa652d/features/support/env.rb#L49). My concern is that somebody might in the future remove the "unused" environment variables from the GitHub action, and in so doing will cause the test to pass when it should fail.
- I'm not sure where this test should live; is `features/sanity_check.yaml` a good home for it, or maybe `features/apps/smokey.yaml`?

## Testing

[^1]: https://github.com/alphagov/smokey/pull/983
[^2]: https://github.com/alphagov/smokey/actions/runs/3099371415/jobs/5018429956

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,11 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+      - run: env ENVIRONMENT="integration" bundle exec cucumber features/apps/smokey.feature
--- a/features/apps/smokey.feature
+++ b/features/apps/smokey.feature
@@ -1,0 +1,9 @@
+Feature: Smokey
+
+  This is a quick sense check, run on every PR against Smokey itself,
+  to verify that nothing is fundamentally broken.
+
+  Scenario: Check that Smokey can make network requests
+    When I request "https://www.gov.uk"
+    Then I should get a 200 status code
+    And I should see "Welcome to GOV.UK"
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -46,10 +46,11 @@ Capybara.register_driver :headless_chrome do |app|
 
   driver = Capybara::Selenium::Driver.new(app, browser_options)
 
+  driver.browser.devtools.send_cmd(
+    'Network.enable'
+  )
+  
   if ENV["RATE_LIMIT_TOKEN"]
-    driver.browser.devtools.send_cmd(
-      'Network.enable'
-    )
     driver.browser.devtools.send_cmd(
       'Network.setExtraHTTPHeaders',
       headers: { 'Rate-Limit-Token': ENV["RATE_LIMIT_TOKEN"] }


features/sanity_check.feature
```diff
@@ -0,0 +1,5 @@
+Feature: Sanity Check
+  Scenario: Check that Smokey is not fundamentally broken
```
ChrisBAshton, line 2: I like the intent of this test. We could be more specific about what it's actually checking though, e.g.

```suggestion
  Scenario: Check that Smokey can make network requests
```

If we want to add context, we can [use multiple lines under `Feature`](https://cucumber.io/docs/gherkin/reference/#feature) at the top of the file about the intent behind this file.

ChrisBAshton (commented):
Nice work so far ⭐ 

`features/smokey.feature` feels like the best place for this sense check (which is a slightly more inclusive term than sanity check; suggest a rename here).

> The test only fails for the above-mentioned Dependabot PR when the RATE_LIMIT_TOKEN environment variable is set (to any value), presumably because of the conditional [here](https://github.com/alphagov/smokey/blob/02784abbbcefe9e7366cf7b93d5f1937bdaa652d/features/support/env.rb#L49). My concern is that somebody might in the future remove the "unused" environment variables from the GitHub action, and in so doing will cause the test to pass when it should fail.

Well done for checking this against an example PR that caused issues 👍 
Looks like [`Network.enable` enables network tracking](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-enable), which is [necessary for setting request headers](https://stackoverflow.com/questions/15645093/setting-request-headers-in-selenium). It does not seem to be, as I first guessed, to enable networking in general (if it were, I'd be surprised that your test still passes when the ENV var is missing!).

I still don't know why whether this block of code executes or not affects your test 🤔 For important gotchas such as this, feel free to add a comment above the `bundle exec` statement along the lines of "# RATE_LIMIT_TOKEN must be set for this test to work properly". But I'd want us to do a little bit of digging to figure out exactly why that's needed first - there may be an opportunity to refactor instead.

robinjam:
Thanks for the review, good points all round!

> sense check (which is a slightly more inclusive term than sanity check; suggest a rename here).

Good catch, I hadn't considered how problematic that term is.

> I still don't know why whether this block of code executes or not affects your test

Based on the stack trace from the failed run[^1] it looks like the `browser` function call in `driver.browser.devtools.send_cmd`[^2] is `require`ing a missing library. If the conditional is skipped then this function never gets called (from this one test at least), and the missing library never gets `require`d.

[^1]: https://github.com/alphagov/smokey/actions/runs/3099371415/jobs/5018429956
[^2]: https://github.com/alphagov/smokey/blob/02784abbbcefe9e7366cf7b93d5f1937bdaa652d/features/support/env.rb#L50

robinjam:
Good shout, I've done that now (and confirmed that the broken PR still fails this check[^1]).

[^1]: https://github.com/alphagov/smokey/actions/runs/3129568728/jobs/5078822772

ChrisBAshton (approved):
Nice one ⭐

PR #1048, opened 2022-11-03
Add test for passing of Rate-Limit-Token request header

ChrisBAshton:
We have some logic to add a `Rate-Limit-Token` request header when a particular environment variable is present: https://github.com/alphagov/smokey/blob/22072e261b38da114015a81f8cd73392331d74a5/features/support/env.rb#L49-L58

Prior to this commit, there were no tests in place to prevent the functionality from being removed or accidentally broken. It's not hugely clear from the PR where it was added (#940) why the feature was introduced in the first place, so it's also not clear where we might see breakages were this code to be removed.

Assuming it is important, this is a quick sense-check Smokey CI test that ensures that the request header is sent when the `RATE_LIMIT_TOKEN` ENV variable is set. This frees us up to refactor Smokey's networking with more confidence.

## Testing

### DIFF
--- a/features/apps/smokey.feature
+++ b/features/apps/smokey.feature
@@ -7,3 +7,11 @@ Feature: Smokey
     When I request "https://www.gov.uk"
     Then I should get a 200 status code
     And I should see "Welcome to GOV.UK"
+
+  Scenario: Check that Smokey passes Rate Limit Token when set
+    Given the 'RATE_LIMIT_TOKEN' ENV variable is set
+    Then any request I make should include the 'Rate-Limit-Token' header
+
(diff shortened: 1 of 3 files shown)


KludgeKML:
Context on when it was added - we were getting rid of the proxy, and needed to introduce the ability to set a rate limit token to bypass the rate limiter that had recently been added to feedback. So I'd expect if this was removed/broken what we'd see is smokey tests starting to intermittently fail for Feedback as requests from the smokey machine get throttled.

ChrisBAshton:
Closed in favour of #1049. (More helpful to see all the changes in context)

PR #1056, opened 2022-11-25
Fix container image build.

sengi:
- Don't try to run `rake webdrivers:chromedriver:update`, since that task and indeed the entire Rakefile were removed in https://github.com/alphagov/smokey/pull/893. Fixes the image build, which had been broken since https://github.com/alphagov/smokey/pull/1049.
- Actually install ChromeDriver, which had been missing since https://github.com/alphagov/smokey/pull/1049.

Other fixes and cleanup while we're there:

- Security fix: don't run as root by default.
- Security fix: don't defeat the Chrome sandbox by default.
- Use govuk-ruby-images, which gets rid of a ton of bad-practice `apt-get upgrade` and the like.
- Remove the JRE, as this was only needed for BrowserMob, which we're no longer using.
- Stop using deprecated `apt-key`.
- Update Bundler from 2.1 to 2.3, since 2.1 had some issues with Ruby 3.1.
- General Dockerfile good practices e.g. use COPY and not ADD for copying files locally, set `-o pipefail` when using shell pipelines, use WORKDIR to change/create directories, etc.

Rollout: will need to set `NO_SANDBOX` in govuk-puppet and govuk-docker if it's not already.

### Testing

[Output from test run on Jenkins](https://deploy.integration.publishing.service.gov.uk/job/Smokey/44041/console)

The [instructions](https://github.com/alphagov/smokey/#running-the-test-suite) for running in govuk-docker seem to be incomplete and I wasn't able to follow them. If it's supposed to be possible to run this in govuk-docker, let me know what I've missed and I can fix the README.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,37 +1,45 @@
-FROM ruby:3.1.2
-RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential \
-  libpq-dev libxml2-dev libxslt1-dev dumb-init default-jre
+ARG base_image=ghcr.io/alphagov/govuk-ruby-base:3.1.2
+ARG builder_image=ghcr.io/alphagov/govuk-ruby-builder:3.1.2
 
-ENV APP_HOME /smokey
-RUN mkdir $APP_HOME
-
-ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE 1
-
-# Install Google Chrome
-RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -
-RUN sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list'
-RUN apt-get update && apt-get install -y google-chrome-stable
+FROM $builder_image AS builder
 
 WORKDIR $APP_HOME
-ADD Gemfile* $APP_HOME/
-ADD .ruby-version $APP_HOME/
+COPY Gemfile* .ruby-version ./
 RUN bundle install
+COPY . ./
 
-ADD . $APP_HOME
 
-RUN bundle exec rake webdrivers:chromedriver:update
+FROM $base_image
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
The changes seem sensible to me 👍 
Also did a manual run of this branch and it [passed](https://deploy.integration.publishing.service.gov.uk/job/Smokey/44043/console).

sengi:
Thanks!

Dockerfile
```diff
@@ -1,37 +1,45 @@
...
+    gpg --dearmor < /tmp/google.pub > "${keyring}" && \
+    echo "deb [arch=${arch} signed-by=${keyring}] https://dl.google.com/linux/chrome/deb/ stable main" \
+        > /etc/apt/sources.list.d/google.list && \
+    install_packages dumb-init google-chrome-stable unzip && \
+    rm -fr /tmp/*
+RUN chrome_ver=$(google-chrome --version | grep -Po '\d+\.\d+\.\d+') && \
+    chromedriver_ver=$(curl -LSfs "${chromedriver_url}LATEST_RELEASE_${chrome_ver}") && \
+    curl -LSfs "${chromedriver_url}${chromedriver_ver}/chromedriver_linux64.zip" \
+        | funzip >/usr/bin/chromedriver && \
+    chmod 755 /usr/bin/chromedriver
```
ollietreend, line 32: I think _all of this gumpf_ is the reason why I originally switched over to Chromium rather than Google Chrome 😅 

Having Chromium in the apt package manager made it as simple as `apt install chromium chromium-driver`, which felt like a much less noisy way to do things.

However: this is a necessary evil if we want/need to be testing against Google Chrome 👍🏻

sengi: Yeah 💯 — tbh I would have just switched it back to Chromium had I not seen that we'd only recently switch to Chrome in #1049 😅 Certainly took me a while to script that ChromeDriver installation, but hopefully it's a more robust outcome overall.

sengi: I suspect we may have to switch back at some point in order to support arm64, but with any luck this should tide us over until everything's on k8s.

PR #1057, opened 2022-11-25
Have headless Chrome use /tmp instead of shared memory.

sengi:
This allows running the WebDriver tests in a container without needing to tune sysctls on the node and reserve a ton of RAM.

Also disable GPU acceleration and the ability to load Chrome extensions while we're at it, since we don't need/want either of those.


ChrisBAshton (approved):
[Passed](https://deploy.integration.publishing.service.gov.uk/job/Smokey/44045/console) 👍

ollietreend (approved):

PR #1074, opened 2023-01-05
Fix container image build.

sengi:
Avoid trying to use `gpg` in the final stage; it's not installed in the govuk-ruby-images base image and we don't actually need it there.

Also some minor cleanup / improvements:
- Omit the `bundle exec` from the default command, now that it's no longer needed.
- Add some files to `.dockerignore`.
- Log the base URLs from Plek that we browse to in the tests, so that it's much easier to debug when a name resolution fails.

Rollout: depends on https://github.com/alphagov/govuk-ruby-images/pull/30

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,12 @@
+*.log
 .git
 .github
-bug.log
+Dockerfile
+Jenkinsfile
+README.md
+deploy.sh
 docs
+jenkins.sh
 log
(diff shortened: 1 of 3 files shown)


sengi:
The failing check is because Jenkins is out of disk space, apparently.

nsabri1 (commented):
OK

nsabri1 (approved):

PR #1092, opened 2023-02-07
Use the default allocator instead of jemalloc.

sengi:
Chromedriver seems to be having a problem since we switched the base image to jemalloc. Let's see if it's causal.

This only affects the k8s environments.


theseanything (approved):

sengi:
Yup, that seems to be it 🤷

Issue #1080, opened 2023-01-16
Confusing/misleading error messages on test failures.


sengi:
There's something about the way Sentry is configured in Smokey that leads to misleading error logs like this when a test fails.

In this example, a network request to the application under test has timed out. From the error messages, it'd be easy to think that the test is broken because of something to do with Sentry, but this isn't the case.

This isn't an isolated example; pretty much any test failure that results in an exception in the test runner looks similar to this.

```
  @app-publishing-api
  Scenario: Can log in to imminence           # features/apps/imminence.feature:4
    When I go to the "imminence" landing page # features/step_definitions/smokey_steps.rb:46
      application_external_url(imminence) = 'https://imminence.integration.publishing.service.gov.uk/admin'
      Net::ReadTimeout with #<TCPSocket:(closed)> (Net::ReadTimeout)
      ./features/support/visiting_pages.rb:7:in `visit_path'
      ./features/step_definitions/smokey_steps.rb:47:in `/^I go to the "([^"]*)" landing page$/'
      features/apps/imminence.feature:5:in `I go to the "imminence" landing page'
    And I try to login as a user              # features/step_definitions/signon_steps.rb:1
    And I go to the "imminence" landing page  # features/step_definitions/smokey_steps.rb:46
    Then I should see "GOV.UK Imminence"      # features/step_definitions/smokey_steps.rb:160
    And I should see "Sign out"               # features/step_definitions/smokey_steps.rb:160
    And I should see "All services"           # features/step_definitions/smokey_steps.rb:160
E, [2023-01-16T13:03:51.781156 #26604] ERROR -- sentry: ** [Sentry] Event sending failed: undefined method `underscore' for "ReadTimeout":String
I, [2023-01-16T13:03:51.781423 #26604]  INFO -- sentry: ** [Sentry] Unreported Event: features/apps/imminence.feature

Net::ReadTimeout
E, [2023-01-16T13:03:51.781475 #26604] ERROR -- sentry: ** [Sentry] exception happened in background worker: undefined method `underscore' for "ReadTimeout":String
```

sengi:
Thanks for fixing this! ❤️

PR #1096, opened 2023-02-17
Remove hint from Sentry event

MuriloDalRi:
The addition of a 'hint' to the Sentry event causes the SDK to fail when sending the event.

This should fix https://github.com/alphagov/smokey/issues/1080

https://trello.com/c/asECCyWk/3096-fix-smokey-sentry-integration-3

### DIFF
--- a/features/support/hooks.rb
+++ b/features/support/hooks.rb
@@ -42,18 +42,16 @@ def capture_error(scenario, exception, js: false)
 
     # Adding the scenario supports further drill down using the
     # Sentry "tags" view. Substituting " is necessary for search.
-    scope.set_tags('cucumber.scenario': scenario.name.gsub('"', "'"))
-    scope.set_tags('cucumber.js_error': js)
-
     # Using the feature (file) as the fingerprint means we only
     # get one issue per feature, keeping the dashboard simple.
-    # The hint is used by govuk_app_config, which will emit stats
-    # for the error class, in addition to what we get in Sentry.
-    Sentry.capture_message(
+    GovukError.notify(
       message,
       fingerprint: [scenario.location.file],
       backtrace: exception.backtrace,
-      hint: { exception: exception }
+      tags: {
+        'cucumber.scenario': scenario.name.gsub('"', "'"),
+        'cucumber.js_error': js
+      }
     )
   end
 end


features/support/hooks.rb
```diff
@@ -47,13 +47,10 @@ def capture_error(scenario, exception, js: false)
 
     # Using the feature (file) as the fingerprint means we only
     # get one issue per feature, keeping the dashboard simple.
-    # The hint is used by govuk_app_config, which will emit stats
-    # for the error class, in addition to what we get in Sentry.
     Sentry.capture_message(
       message,
       fingerprint: [scenario.location.file],
       backtrace: exception.backtrace,
-      hint: { exception: exception }
```
ChrisBAshton, line 56: Without this, we'll lose the [statsd error counting](https://github.com/alphagov/govuk_app_config/blob/87e445eccee5fba2449a170d5ba628e8a380fcb8/lib/govuk_app_config/govuk_error/configuration.rb#L107-L115), but I don't think we intended to use that anyway, so it's probably fine.

Can you add more detail about _why_ it fails? (The [commit where it's added](https://github.com/alphagov/smokey/pull/931/commits/a3c2836458edb537864f755d7fa3a9b1d46d68cd) references code in [sentry-ruby/lib/sentry/hub.rb](https://github.com/getsentry/sentry-ruby/blob/7848d59450ae6bcb8dea6c141abdcdfe7b218567/sentry-ruby/lib/sentry/hub.rb#L100) but it doesn't look like it's changed much since then).

Is there an error message or does it just fail to work? Can you link to a Sentry issue/error that proves this change works? etc

Also I'm not sure why we refer to `Sentry` directly, rather than `GovukError`, which is the [documented approach in govuk_app_config](https://docs.publishing.service.gov.uk/repos/govuk_app_config.html#error-reporting). Perhaps this wouldn't have broken if we'd have used that as an abstraction layer, since that's kept up to date? Could be worth looking into.

ChrisBAshton (commented):
If this fixes it, nice sleuthing! 🕵️ 
Though I have a few questions first.

features/support/hooks.rb
MuriloDalRi: So I'm not sure this has ever worked, the error message is the same as on the issue linked, not very helpful.

You can see the errors [here](https://govuk.sentry.io/issues/?environment=development&project=6370326&statsPeriod=7d) from my local run.

I tested calling GovukError instead of Sentry directly and the results are exactly the same, it works only without the hint. I also thought it was weird we're calling Sentry directly so I'll change it to use GovukError instead.

ChrisBAshton (approved):
LGTM 🎉 

I'd suggest deploying this branch first and do a Smokey run to make sure it a) doesn't break Smokey and b) errors start appearing in Sentry. Then it should be good to merge. Thanks!

MuriloDalRi:
Tested in integration and [errors appeared in Sentry](https://govuk.sentry.io/issues/?environment=integration-blue-aws&project=6370326&statsPeriod=24h)

PR #1100, opened 2023-02-27
Add tests for sidekiq-monitoring

brucebolt:
These ensure we are able to load the dashboard for the sidekiq monitoring of each application where it is used.

This has been tested in https://deploy.integration.publishing.service.gov.uk/job/Smokey/49158/console.

[Trello card](https://trello.com/c/aVx282Rt)


BeckaL (approved):
Looks good to me, thanks for linking to the smokey run. No more accidental breakages from now 🥳

PR #1101, opened 2023-03-01
Don't probe sidekiq dashboard app on k8s.

sengi:
We're not running the [sidekiq-monitoring dashboard app](https://github.com/alphagov/sidekiq-monitoring) in k8s.

Test run passed on the integration cluster.


nsabri1 (approved):
OK

PR #1119, opened 2023-04-14
Update manual test instructions for EKS

ChrisBAshton:
We will soon be [removing Smokey from Jenkins](https://trello.com/c/ahTC4Wjj/3134-disable-smokey-in-jenkins-2) entirely (as it only tests the EC2 versions of apps, which no longer serve traffic). We therefore need to update our processes for testing Dependabot PRs.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -1,17 +1,15 @@
 # Deployment
 
-## Before you merge
-
 This app only has minimal CI set up, running a single check to ensure that Smokey is able to make requests. It would be misleading to run all of the tests on CI because their behaviour changes depending on the environment they are run in. You should manually test in applicable environments, until you are confident your change is not a breaking one.
 
 Example steps to test in Integration:
 
-- Click "Configure" for the [Smokey job][]
-- Change the "Branch specifier" to the name of your branch
-- Run a new build of the Smokey project and check the results
+- Navigate to the [Deploy GitHub Action](https://github.com/alphagov/smokey/actions/workflows/deploy.yml) tab on the repo
+- Click “Run workflow” and enter the name of the branch under "Commit, tag or branch name to deploy". (NB: leave the "Use workflow from" option set as main)
+- Wait for the ["sync status" in Argo CD](https://argo.eks.integration.govuk.digital/applications/cluster-services/smokey) to update (it will take several minutes)
+- Wait for a Smokey run to be triggered (happens every 10 minutes), or trigger one through GitHub Actions.
+- Check the results from the logs
 
-## After you merge
+If the build passes, you can merge, and Argo CD will automatically update itself back to `main`.
 
-The manual [Smokey job][] (in every Jenkins environment) will pick up changes on `main` automatically.
-
-[Smokey job]: https://deploy.integration.publishing.service.gov.uk/job/Smokey/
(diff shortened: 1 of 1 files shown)


docs/deployment.md
```diff
@@ -1,17 +1,15 @@
...
 This app only has minimal CI set up, running a single check to ensure that Smokey is able to make requests. It would be misleading to run all of the tests on CI because their behaviour changes depending on the environment they are run in. You should manually test in applicable environments, until you are confident your change is not a breaking one.
 
 Example steps to test in Integration:
 
-- Click "Configure" for the [Smokey job][]
-- Change the "Branch specifier" to the name of your branch
-- Run a new build of the Smokey project and check the results
-
-## After you merge
+- Navigate to the [Actions](https://github.com/alphagov/smokey/actions) tab on the repo
```
theseanything, line 7:
```suggestion
- Navigate to the [Deploy GitHub Action](https://github.com/alphagov/smokey/actions/workflows/deploy.yml) tab on the repo
```

theseanything: You need to be on the specific workflow before you get given the "Run workflow" button.

docs/deployment.md
```diff
@@ -1,17 +1,15 @@
...
+- Click “Run workflow” and enter the name of the branch under "Commit, tag or branch name to deploy". (NB: leave the "Use workflow from" option set as main)
+- Wait for the ["sync status" in Argo CD](https://argo.eks.integration.govuk.digital/applications/cluster-services/smokey) to update (it will take several minutes)
+- Wait for a Smokey run to be triggered (happens every 10 minutes), or trigger one through GitHub Actions.
```
theseanything, line 10: optional: could tell people to create k8s job from a cronjob, so you don't have to wait:

```
k -n apps create job --from=cronjob/smokey smokey-test-$USER-$(date +"%Y-%m-%d-%H-%M")
```

docs/deployment.md
```diff
@@ -1,17 +1,15 @@
...
+- Check the results
```
theseanything, line 11: optional: 

```suggestion
- Check the results from the logs
```
You can also use command like: 
```
k logs -n apps jobs/smokey-test-$USER-$DATE
```

docs/deployment.md
```diff
@@ -1,17 +1,15 @@
...
 
-The manual [Smokey job][] (in every Jenkins environment) will pick up changes on `main` automatically.
+If the build passes, you can merge, and Argo CD will automatically update itself back to `main`.
```
theseanything, line 13: nit: might be simpler to say:

```suggestion
If the build passes, you can merge, and your changes will automatically be deployed.
```

theseanything (approved):
Thanks for updating these! Couple comments -

docs/deployment.md
ChrisBAshton: Wouldn't it be easier to get people to visit https://github.com/alphagov/smokey/actions/workflows/ci.yml and click "Run workflow"? Or does that do something subtly different?

docs/deployment.md
ChrisBAshton: I want to make it explicit that Argo will end up back in a stable state, no further action needed from the user. I've probably worded it poorly though.

PR #1130, opened 2023-05-18
Stop grouping error reports in Sentry

robinjam:
[Trello card](https://trello.com/c/Eb4Nvyjg/3172-audit-and-identify-flakey-smokey-tests-3)

Currently, the events we send to Sentry are grouped by feature. This means that if a feature contains several flakey tests then the failures all get jumbled together and the only way to see them all is to click back/forward inside Sentry dozens/hundreds of times until you find something new.

## Testing


robinjam:
Nevermind, I didn't realise you could drill down by tag 🤦‍♂️

PR #1135, opened 2023-06-07
Update/Remove licence-finder tests

KludgeKML:
Licence Finder is being retired, and licence URLs are going to update (although redirects will be in place for them).

- Remove licence finder test (it no longer exists)
- Update URL in frontend test to check it can talk to licensing so that it doesn't go through the redirect.

https://trello.com/c/VWSPy1u4/2055-create-pr-updating-smokey-test

Has been tested locally against integration - the updated test as expected fails when the new licence finder licence urls are not present, but passes when the licence data import task has been run.

## Testing


leenagupte (approved):
👍

PR #1143, opened 2023-06-27
Move CSV Preview test from Whitehall to Frontend

brucebolt:
CSV Previews are now rendered by Frontend, instead of Asset Manager, so we need to test that application instead.

Depends on https://github.com/alphagov/govuk-helm-charts/pull/1186.

[Trello card](https://trello.com/c/1kmt0VHe)


aldavidson (approved):
Looks pretty straightforward :+1:

PR #1144, opened 2023-06-29
Update README etc. now that we no longer run on Jenkins.

sengi:
- Remove obsolete script for running under Jenkins (which we haven't for some time now).
- Update the README to reflect the move to Kubernetes (plus some trivial house style fixes)


ChrisBAshton (approved):
LGTM 👍

Nitpick: this PR moves to hard-wrapping long lines onto multiple lines, which isn't an issue but also isn't necessary (and creates some unnecessary diffs here, making it more cumbersome to see the 'true' commit that introduced a given line of documentation). There isn't an official house style AFAIK - we're pretty inconsistent throughout our docs - but it's [come up before](https://github.com/alphagov/local-links-manager/pull/786#issuecomment-820357427) and the general view was that it's more important that there's a consistency of style within the file, rather than globally adopting one style or the other. Just something to bear in mind for future PRs.

sengi (commented):
Sure — I'll see if I can configure my editor to leave the long lines alone next time (or if not then I'll reformat in a separate commit first). Right now it's set up to pass markdownlint (which I know we don't have yet).

sengi:
Oh btw just in case you haven't come across it, GitHub has an ignore-whitespace option in the diff view. It's hidden in dropdown revealed by the cog icon next to "File filter" and "Conversations".

PR #1154, opened 2023-07-20
Reapply licence smoke test in integration

1pretz1:
We disabled the busking-licence test as there was datasync issues, however this seems to be resolved now so we can reapply the test in integration.

https://www.integration.publishing.service.gov.uk/find-licences/busking-licence

## Testing


KludgeKML (approved):
LGTM

PR #1180, opened 2023-09-12
Update search result selector

sihugh:
We don't have `a` tags nested in each other. This means the test can't find the search result to click on.

The search result markup looks like this (trimmed for brevity):

```
<div class="gem-c-document-list__item-title">
  <a class="  govuk-link" href="/pay-dartford-crossing-charge">Pay the Dartford Crossing charge (Dart Charge)</a>
</div>
```

so removing the first `a`, should fix things.


robinjam (approved):
Will need rebasing on main (I've merged my PR now), but LGTM 👍

PR #1197, opened 2023-10-27
Add comment about GA4 to user-agent lines

AshGDS:
## What
Adds a comment about how the user agent is now being used by GA4.

## Why
So that devs are aware if they want to change the value.

## Testing


andysellick (approved):
🥇

PR #1214, opened 2023-12-15
Debug smokey

minhngocd:
Figure out why tests are failing

## Testing


theseanything:
We don't want debug logic in production code.

PR #1225, opened 2024-01-08
Remove the search-analytics smokey tests

leenagupte:
The three search queries used in these smokey tests are inflating and polluting the clickboost search data.

The GA4/Analytics team tried to filter out the Smokey data at the GA4 end but weren't fully able to.

The test can be removed for a few reasons:

1. The clickboost data is being used to evaluate the new search product and that can't happen effectively if the clickboost data thinks pages are popular based on clicks occurring in tests. Once the new product is live it will continue to effect the results that users see.
2. search-analytics hasn't worked properly since the move to kubernetes, though work is being done to retify this.
3. These tests check for Univeral Analytics which is being removed, so the tests will need to br written anyhow.

## Testing

Unrelated Feedback test fails in [integration](https://argo.eks.integration.govuk.digital/applications/cluster-services/smokey?orphaned=false&node=batch%2FJob%2Fapps%2Fsmokey-28412250&tab=logs&resource=).

### DIFF
--- a/features/apps/finder_frontend.feature
+++ b/features/apps/finder_frontend.feature
@@ -19,12 +19,6 @@ Feature: Finder Frontend
     When I search for "<keywords>"
     Then I should see some search results
     And the search results should be unique
-    And search analytics for "<keywords>" are reported
-    When I go to the next page
-    Then the "contentsClicked" event is reported
-    When I click result 1
-    Then the "navFinderLinkClicked" event for result Search.1 is reported
-    And the "UX" event for result click is reported
 
     Examples:
     | keywords         |
--- a/features/step_definitions/analytics.rb
+++ b/features/step_definitions/analytics.rb
@@ -1,18 +1,3 @@
-Then /^search analytics for "(.*)" are reported$/ do |term|
-  sought = CGI::escape("/search/all?keywords=#{term.sub(' ', '+')}")
-  expect(browser_has_analytics_request_containing sought).to be(true)
-end
-
-Then /^the "(.*)" event is reported$/ do |event|
-  sought = "ec=#{event}"
-  expect(browser_has_analytics_request_containing sought).to be(true)
-end
-
-Then /^the "(.*)" event for result (.*) is reported$/ do |event, n|
-  sought = "ec=#{event}&ea=#{n}"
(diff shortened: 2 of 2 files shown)


robinjam (approved):

features/apps/finder_frontend.feature
```diff
@@ -19,12 +19,6 @@ Feature: Finder Frontend
     When I search for "<keywords>"
     Then I should see some search results
     And the search results should be unique
-    And search analytics for "<keywords>" are reported
-    When I go to the next page
-    Then the "contentsClicked" event is reported
-    When I click result 1
-    Then the "navFinderLinkClicked" event for result Search.1 is reported
-    And the "UX" event for result click is reported
```
gclssvglx, line 27: Bit of a 'meta' question sorry - does the fact that these tests actually create analytics data mean we should never have/use/write tests like this? Surely, we're always polluting data and unless that can be subsequently ignored it'll remain an issue. 

Could we add/change something into the analytics data being created here - 'TEST_PLEASE_IGNORE' for example - so that it can be reliably be ignored.

As a general thought, I wonder if this is the best way to test that analytics data is being fired and captured 🤔 

Do we know if it's specific keywords that cause the pollution or just that fact that anything in there is an issue? Just thinking if we can exclude tests based on specific keywords.

Sorry, a lot of questions...

gclssvglx (approved):
Great stuff @leenagupte 👍 

More than happy to approve this as these tests clearly pollute our 'real' analytics data - clearly undesirable - so needs to be removed. I've added some thoughts and questions but I feel/know some or all of these are being addressed by the GA4 project, so can probably be treated as rhetorical.

features/apps/finder_frontend.feature
leenagupte: It turns out the clickboost score is being distorted. The LTR model in the old search-api has added logic to mitigate this, but the new search is just using the clickboost scores as is, because why wouldn't you? 😱 

The test data just made it easier to spot. One of the things that the PA stumbled on was that the order of results for "passport" looked odd.

I do wonder whether checking that the analytics are being reported need to be a smokey test at all, or yes, should have someway of identifying itself as test data that should be ignored.

PR #1228, opened 2024-01-09
Fix static feature

unoduetre:
# What

Form actions can contain a domain name, but `feedback_steps.rb` does 
not take it into account and it assumes the action URL can only contain a path but not a domain. This PR fixes that by checking if the action url **ends with** (as opposed to be equal to) the path.

# Why

This is a proposed PR fixing an ongoing bug with the smokey test. It does not introduce new any functionalities. For this reason it does not contain the ticket reference.

## Testing


robinjam (approved):

unoduetre (commented):
Looks good to me

PR #1242, opened 2024-01-29
Update collections.feature test to test new stable page

georges1996:
## Testing

- This example page pulls data in from the `Services.search_api.search` in the collections app https://www.integration.publishing.service.gov.uk/world/afghanistan/news
-  The `app/models/world_location_news.rb` file makes a number of calls to this api pulling in different collections here https://github.com/alphagov/collections/blob/main/app/models/world_location_news.rb#L69-L83. 
- Here is where the call is made `app/services/search_documents.rb`. 
- When testing locally and removing these calls we don't see the various sections on this page https://www.integration.publishing.service.gov.uk/world/afghanistan/news

Ticket: https://trello.com/c/nGHmbA30/2332-update-smokey-before-we-retire-hmrc-services-and-information-page-l

### DIFF
--- a/features/step_definitions/smokey_steps.rb
+++ b/features/step_definitions/smokey_steps.rb
@@ -185,10 +185,13 @@ Then /^JSON is returned$/ do
   expect(JSON.parse(@response.body).class).to eq(Hash)
 end
 
-When /^I see links to pages per topic$/ do
-  pages = Nokogiri::HTML.parse(page.body).css(".browse-container a")
-  unless pages.any?
-    fail "There are no links on this Services and Information page"
+When /^I see the links pulled form search_api$/ do
+  latest_section = Nokogiri::HTML.parse(page.body).css("#latest")
+  latest_items_list = latest_section.css(".gem-c-document-list ul")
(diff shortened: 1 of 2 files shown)


features/step_definitions/smokey_steps.rb
```diff
@@ -185,9 +185,12 @@ def should_see(text)
...
 
-When /^I see links to pages per topic$/ do
-  pages = Nokogiri::HTML.parse(page.body).css(".browse-container a")
-  unless pages.any?
+When /^I see the links pulled form search_api$/ do
+  latest_section = Nokogiri::HTML.parse(page.body).css("#latest")
+  latest_items_list = latest_section.css(".gem-c-document-list ul")
+  latest_items = latest_items_list.css(".gem-c-document-list__item-metadata")
+
+  unless latest_items.any?
```
hannako, line 193: Shall we change the error here: `fail "There are no links on this Services and Information page"`

Tetrino: I agree, this seems a bit vague and should tell us what we're expecting directly - if only because sometimes we have many of these failures and need a quick reference.

Tetrino (approved):

PR #1246, opened 2024-02-01
Fix postcode used to test POST request for CDN

theseanything:
Updated the postcode to our current office, previous postcode didn't return any results.


samsimpson1 (approved):

sengi:
Curious that we stopped returning results for the previous postcode though 🤔

PR #1261, opened 2024-02-21
Rename licensify tests

theseanything:
This renames the test label for licensify so that they are correct referenced in the post sync workflow, which expected them to match the ArgoCD App name.


sengi (approved):
Good find 😅

PR #1312, opened 2024-07-09
Remove uniqueness test for search results

KludgeKML:
- The finder-frontend scenario these are used in is "Check the frontend can talk to Search API", and that doesn't really need to prove uniqueness, only that _any_ results come back.
- The uniqueness test really seems to be about search _quality_, and the search team are already monitoring quality much more intensively than this simple test can.
- The tests as they are are a little fragile (they break if the result doesn't have descriptive text), and although they could easily be re-written to be less fragile, it's not clear they actually add anything.

## Testing

### DIFF
--- a/features/apps/finder_frontend.feature
+++ b/features/apps/finder_frontend.feature
@@ -18,7 +18,6 @@ Feature: Finder Frontend
     Given I consent to cookies
     When I search for "<keywords>"
     Then I should see some search results
-    And the search results should be unique
 
     Examples:
     | keywords         |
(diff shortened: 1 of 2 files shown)


csutter (approved):
Looks good and makes sense ✅ 

While we're at it - is it worth improving the name of this scenario ("_check the frontend can talk to Search API_")?

Just thinking that it _technically_ checks for Search API v2, but also that's an implementation detail and really what we're interested in is something along the lines of "check some search results are returned"...

KludgeKML:
I think we probably shouldn't add versions here. By design, IIRC, a lot of these tests are specifically like this (ie they are actually just checking if two components that should be able to talk can), so I think we should probably leave it as is at the moment to stay coherent with the other tests.

PR #1313, opened 2024-07-09
Remove dangling reference to a deleted Analytics step.

sengi:
Since #1310, `cdn.feature` has been failing with:

```
Undefined Scenarios:
cucumber -p production features/cdn.feature:9 # Scenario: Check an A/B test is persistent
...
You can implement step definitions for undefined steps with these snippets:
Then('the bucket is reported to Google Analytics') do
...
```


KludgeKML (approved):
LGTM - thanks, not sure how I missed that one.

PR #1332, opened 2024-09-23
Add tests for GOV.UK Chat

jackbot:
https://trello.com/c/bnG6AjtP/1937

Adds two simple tests:

1. For a non-signed in user viewing a static page. We use the about page here instead of the homepage as the Chat app can be turned on and off to allow and disallow public access while it's still in beta mode. So the homepage can change state, which is why we test against a static page that won't change.
2. For a signed-in user viewing the admin area

Chat is set up slightly differently on integration/staging vs prod currently. Chat is still in beta so it's only available via the external URL on prod.  So a custom step definition was needed here to set the external URL differently on integration. Once we've got all environments consistent, we can delete this whole step definition file and just use the existing "When I visit x" definition.

### DIFF
--- a/features/apps/govuk_chat.feature
+++ b/features/apps/govuk_chat.feature
@@ -1,0 +1,12 @@
+@app-govuk-chat
+Feature: GOV.UK Chat
+  Scenario: Can view a static page
+    When I visit the GOV.UK Chat about page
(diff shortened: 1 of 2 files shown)


davidgisbey (approved):
looks good to me

